PR #293, opened 2015-01-07
Certificate info view controller

chrisballinger:
Seems like this code is salvageable. Probably easier to just rip out the standalone classes and reintegrate than to fix the merge conflicts because of the project restructuring.


chrisballinger:
@davidchiles bump

PR #312, opened 2015-01-29
Merge pull request #2 from ChatSecure/master

dianaafanador3:
Updates from main project


davidchiles:
This appears like this pull request is adding back code that was removed a while back. It's not clear from the comments what's in this pull request.

PR #338, opened 2015-04-14
Hold to record Audio

davidchiles:
This changes the UI for recording audio from a separate UIViewController to buttons and UI inside a subclass of OTRMessagesViewController.


chrisballinger:
"Hold to talk" view on top of text area still doesn't get removed when view controller is popped from the stack. Steps to reproduce:

1. Bring up "hold to talk" view by pressing microphone
2. Press back to "Chats"
3. Go back into conversation ("Hold to talk" view is still there)
4. Try toggling with the microphone button and then trying to hold to talk again

chrisballinger:
Looks great! :+1:

PR #366, opened 2015-08-13
XEP-0198

davidchiles:
Support for XEP-0198. Not sure what the best parameters are for 'timeouts' right now we request an ack every 10 messages or 90 seconds. It could probably be more frequent without much downside.


ChatSecure/Classes/Controllers/XMPP/OTRStreamManagementYapStorage.m
```diff
@@ -0,0 +1,312 @@
...
+@property (nonatomic) dispatch_queue_t parentQueue;
+
+@end
+
+@implementation OTRStreamManagementYapStorage
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+        self.databaseConnection = [[OTRDatabaseManager sharedInstance] newConnection];
```
chrisballinger: @davidchiles Is it possible to somehow pass in the databaseConnection as an initialization parameter, or a set property?

chrisballinger:
Overall looks really good! Just added 1 small comment. It there a good way to write some small tests for this new class, or is it too tightly coupled to XMPPFramework?

PR #408, opened 2015-11-13
Develop extensibility

N-Pex:
Changes to allow extending onboarding and attachment picking.

### DIFF
--- a/ChatSecure/Classes/OTRAppDelegate.h
+++ b/ChatSecure/Classes/OTRAppDelegate.h
@@ -50,6 +50,9 @@
 /** Override this in subclass to use a different theme class */
 - (Class) themeClass;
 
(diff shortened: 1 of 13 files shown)


ChatSecure/Classes/OTRAppDelegate.h
```diff
@@ -50,4 +50,10 @@
 /** Override this in subclass to use a different theme class */
 - (Class) themeClass;
 
+/** Override this in subclass to use a different conversation view controller class */
+- (Class) conversationViewControllerClass;
```
chrisballinger: @N-Pex I think we should move these to the `OTRTheme` class instead of on the app delegate

chrisballinger:
Looks good! Thank you!

PR #446, opened 2016-03-01
Add knock UI with 'cache' object

davidchiles:
I decied to add a 'state' object. I'm not super pleased with it because it's kinda brittle. But there is so much state to keep track of when changing the 'send' button. between OTRData mode and normal send and knock.


ChatSecure/Classes/Controllers/OTREncryptionManager.m
```diff
@@ -105,6 +105,43 @@ - (OTRProtocolType)prototcolTypeForString:(NSString *)typeString
...
+{
+    if (!queue) {
+        queue = dispatch_get_main_queue();
+    }
+    
+    __block BOOL currentlyTrusted = NO;
+    __block BOOL hasTrustedFingerprints = NO;
+    __block OTRKitMessageState messageState = OTRKitMessageStatePlaintext;
+    
+    dispatch_group_t group = dispatch_group_create();
```
chrisballinger: @davidchiles Good use of a dispatch group

chrisballinger:
@davidchiles I merged the `tor-time` branch into `knock-UI` because the last things that need to be done for Tor accounts is somewhat related:

1. Hide the knock button (may be solved by 2)
2. Disable token exchange (both receive & send)

chrisballinger:
@davidchiles Here are some issues I've noticed from testing:

* If you open the app via a knock and your account isn't set to auto-login, the other contact will appear offline and you'll see the knock button. By default most users will be unaffected so that's good.
* Messages sent from a stale OTR session are marked as delivered but are silently discarded on the other side. I'll make another issue for this.

Otherwise seems to work pretty well!

PR #454, opened 2016-03-17
Added auto restart otr on entering certain buddy conversations

davidchiles:


chrisballinger:
The Strings file is messed up so I can't merge this yet.

We can probably improve this a little further. If your client goes offline, or the resource associated with this OTR session becomes unavailable, the OTR session should be ended. It also doesn't make sense to send OTR requests to offline contacts until they are online.

We could improve this further for Knocks as well. If you're in the chat screen with an offline contact and they become online, we should send an OTR request automatically. This would work well with the message queue where messages are held until an OTR session can be established.

PR #456, opened 2016-03-23
Target pod

davidchiles:
This pull request accomplishes two things.

1. Change to use `target` instead of `link_with` to fix the impeding removal in cocoapods 1.0.0
2. Update all the remaining pods to the latest versions and back to Trunk when possible.


chrisballinger:
Merged into `3.2-beta`

Issue #367, opened 2015-08-14
Support for XEP-0357: Push


iNPUTmice:
Please add support for [XEP-0357: Push](http://xmpp.org/extensions/xep-0357.html). I know you have started working on your own implementation but since there is a XEP for that I believe this would be the better solution. An implementation for ejabberd can be found at [royneary/mod_push](https://github.com/royneary/mod_push)

iNPUTmice:
[![Bountysource](https://www.bountysource.com/badge/issue?issue_id=25760816)](https://www.bountysource.com/issues/25760816-support-for-xep-0357-push?utm_source=25760816&utm_medium=shield&utm_campaign=ISSUE_BADGE)

chrisballinger:
Oh yeah, we are working on that as well. Our push solution is complimentary (acts as the app server) component, and I just need to write a pubsub node to translate the xmpp traffic to my own API. 

To work around the lack of server support, we are allowing people to send pushes directly as well, passing the tokens/endpoint information through a specially defined OTR TLV.

chrisballinger:
We actually would like to work with you to integrate our other push solution for a graceful fallback for users on servers that don't support the new XEP. The Android SDK is almost complete and we will be integrating it into ChatSecure soon

iNPUTmice:
Sorry for the late reply. I originally read your comment on my phone and decided to reply later from my desktop computer.

> We actually would like to work with you to integrate our other push solution for a graceful fallback for users on servers that don't support the new XEP

I understand that you might have customers that need your (proprietary) push solution but generally speaking I think it would be advisable to steer users towards upgrading their servers instead of providing them with an easy way out. Using up-to-date servers will not only give you more features (Stream Management, MAM, carbons, PEP (used by OMEMO)) but also increases security (forced PFS on s2s connctions)

I can think of very few reasons for users not to switch to a proper server.

chrisballinger:
Agreed! Our push solution acts as the "app server" component of XEP-0357 that communicates with APNS/GCM directly. It is just a very simple REST API for provisioning devices and sending raw pushes. Our upcoming XMPP PubSub node will simply translate push requests from an XMPP server to our REST API. Totally complimentary efforts.

[...]

PR #496, opened 2016-05-19
XEP-0357 Support

chrisballinger:
There are still some things that could be improved.

Currently on every stream connection it checks for server support of XEP-0357 and re-registers with a new push token. This is probably too aggressive, but it's the least error prone. This push token is not associated with a buddyKey which may cause issues elsewhere, but seems fine so far.

Another issue is that `mod_cloud_notify` is too aggressive when sending push notifications, and will even send them for typing notifications. Since we display "Someone wants to chat" for every received push that doesn't map to a buddy, it can get a little noisy. We can't control the behavior of other XMPP servers so we should address this in the PubSub node and elsewhere in the client.

We will also want to strip as much content as possible in the PubSub node to avoid leaking extraneous incoming data.

### DIFF
--- a/ChatSecure/Classes/Controllers/PushController.swift
+++ b/ChatSecure/Classes/Controllers/PushController.swift
@@ -122,8 +122,9 @@ public class PushController: NSObject, OTRPushTLVHandlerDelegate, PushController
         
     }
     
-    
-    
+    public func getPubsubEndpoint(completion:(endpoint:String?,error:NSError?) -> Void) {
+        self.apiClient.getPubsubEndpoint(completion)
+    }
     
     
     public func getNewPushToken(buddyKey:String, completion:(token:Token?,error:NSError?) -> Void) {
--- a/ChatSecure/Classes/Controllers/PushController.swift
+++ b/ChatSecure/Classes/Controllers/PushController.swift
@@ -129,7 +129,7 @@ public class PushController: NSObject, OTRPushTLVHandlerDelegate, PushController
     }
     
     
-    public func getNewPushToken(buddyKey:String, completion:(token:TokenContainer?,error:NSError?) -> Void) {
+    public func getNewPushToken(buddyKey:String?, completion:(token:TokenContainer?,error:NSError?) -> Void) {
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)) {[weak self] () -> Void in
             guard let tokenContainer = self?.storage.unusedToken() else {
                 self?.updateUnusedTokenStore({[weak self] (success, error) -> Void in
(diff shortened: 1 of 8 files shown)


chrisballinger:
This PR resolves https://github.com/ChatSecure/ChatSecure-iOS/issues/367

weiss:
First of all: *yay* for implementing this!

> Currently on every stream connection it checks for server support of XEP-0357 and re-registers with a new push token.

Yes, I see no way to get around this in a reliable way.

> Another issue is that `mod_cloud_notify` is too aggressive when sending push notifications, and will even send them for typing notifications. Since we display "Someone wants to chat" for every received push that doesn't map to a buddy, it can get a little noisy. We can't control the behavior of other XMPP servers so we should address this in the PubSub node and elsewhere in the client.

I think enabling [CSI][1] with (if you're using Prosody) [`mod_filter_chatstates`][2] should do the trick.

FWIW, my idea for XEP-0357 support in ejabberd is to basically follow the [business rules suggested by Daniel][3].

[1]: https://xmpp.org/extensions/xep-0352.html
[2]: https://modules.prosody.im/mod_filter_chatstates.html
[3]: http://mail.jabber.org/pipermail/standards/2016-February/030925.html

weiss:
> I think enabling CSI with (if you're using Prosody) `mod_filter_chatstates` should do the trick.

Oh, `mod_cloud_notify` just notifies on offline messages, right?  Then CSI won't help, of course.  (But I thought chat states [aren't supposed to be written to offline storage][1])?

Either way, I think this would best be solved on the XMPP server side; as I thought ideally, your PubSub node wouldn't even see the stanza payload (and would therefore not be able to filter out notifications based on stanza contents).

[1]: https://xmpp.org/extensions/xep-0160.html#types

chrisballinger:
@weiss We've noticed that `mod_cloud_notify` currently doesn't store the push settings across restarts, so re-registering on every login seems to be necessary.

On a side note, I noticed that when checking the server capabilities for the user's JID, it doesn't show support for push, only when you check the server's JID. Perhaps this should be addressed in the spec.

iNPUTmice:
> Currently on every stream connection it checks for server support of XEP-0357 and re-registers with a new push token. This is probably too aggressive, but it's the least error prone

This is probably a good idea anyway so your server can associate a stream with a specific token and take appropriate action like for example not push for that token if the stream is still alive.

I would strongly suggest to read my [business rules](http://mail.jabber.org/pipermail/standards/2016-February/030925.html) and either follow them or if you disagree comment on the mailing list so we can come to an agreement so all clients and servers do more or less the same thing.

iNPUTmice:
> On a side note, I noticed that when checking the server capabilities for the user's JID, it doesn't show support for push, only when you check the server's JID. Perhaps this should be addressed in the spec.

The spec is actually pretty clear. It's prosody that should change the behavior here. (FWIW Conversations is checking both. The server and the accounts disco)

chrisballinger:
Yeah we check for both as well.

PR #498, opened 2016-05-24
Creating an account via nickname ensures JID is unique.

davidchiles:
First it tries to create a JID using the desired nickname. If the
server responds that there is a conflict it appends two random byes in
hex to the end until it is no longer a conflict.

### DIFF
--- a/ChatSecure/Classes/Controllers/Errors.swift
+++ b/ChatSecure/Classes/Controllers/Errors.swift
@@ -8,6 +8,33 @@
 
 import Foundation
 
+public extension NSError {
+    class func XMPPXMLError(error:OTRXMPPXMLError, userInfo:[String:AnyObject]?) -> NSError {
+        return self.chatSecureError(error, userInfo: userInfo)
+    }
+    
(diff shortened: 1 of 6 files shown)


n8fr8:
Does it clean up the nickname, so that if someone enters "I am super cool dude 123!!!" it turns it into iamsupercooldude123 ? That is another aspect of this.

n8fr8:
It can also be done pretty easy with a simple alphanumeric regex filter

chrisballinger:
I believe it has always done this by using stringprep

ChatSecure/Classes/Controllers/Errors.swift
```diff
@@ -41,4 +68,32 @@ extension PushError {
     func error() -> NSError {
         return NSError(domain: kOTRErrorDomain, code: self.rawValue, userInfo: [NSLocalizedDescriptionKey:self.localizedDescription()])
     }
+}
+
+
+@objc public enum OTRXMPPXMLError: Int {
+    case UnkownError   = 1000
```
chrisballinger: @davidchiles Unkown -> Unknown

chrisballinger:
@davidchiles PR Looks good aside from a small typo, hopefully there are no side effects in other parts of the app with the new error format.

PR #520, opened 2016-06-23
Catch errors where the server does not allow non-english chars in JID

davidchiles:
This removes all non english characters using regex, `[^A-Za-z0-9]`. If the jid has no length then we use `[OTRBranding xmppResource]` as the base JID.

As always if a JID conflicts with one on the server then a random string is appended it it. We max out at trying 5 times just to make sure we don't get in an infinite loop.

### DIFF
--- a/ChatSecure.xcodeproj/project.pbxproj
+++ b/ChatSecure.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		3A971C2C5AB08E34FDD18676 /* Pods.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CCD1F8E85C658C8D3865B584 /* Pods.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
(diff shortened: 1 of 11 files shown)


N-Pex:
I see some strangeness happening when trying with some tibetan characters. After inputting these in nickname, I first see a policy error, then trying with "zom" I see a conflict error, then, after random chars are appended the account is registered. However, right after this I get another policy error back. It seems like it tries to register again with faulty original jid?

Below just the last part of my log, after successful registration of "zom.1234":

2016-06-23 12:49:43:160 Zom[32996:4722637] OTRXMPPManager: xmppStreamDidConnect:


2016-06-23 12:49:43:160 Zom[32996:4722637] SEND: <iq type="set"><query xmlns="jabber:iq:register"><username>ངགཧཧཧ</username><password>D2gJnphNM1kf9tQ=</password></query></iq>


2016-06-23 12:49:43:297 Zom[32996:4722216] RECV: <iq xmlns="jabber:client" type="error"><error type="modify"><policy-violation xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/><text xmlns="urn:ietf:params:xml:ns:xmpp-stanzas">Username is blocked</text></error></iq>

PR #526, opened 2016-07-01
Message Queue

davidchiles:
This changes the method for sending messages. The behavior shouldn't change in this pull request. All messages are sent using a `OTRYapMessageSendAction`. 

This does not improve the reliability of sending media messages.

Todo:
- [x] Resend button for messages that failed


chrisballinger:
This is already merged into `omemo`

PR #564, opened 2016-09-28
Send small audio files from memory

N-Pex:
The limit a bit arbitrarily set to 1Mb.


chrisballinger:
Not a bad idea for a fix, but it's bothersome that we still haven't identified the root cause. It's likely either an error or race condition in the `copyDataFromFilePath` or some problem with GCDWebServer..

PR #565, opened 2016-09-29
Check that super implements selector

N-Pex:
Fix crash when super does not implement this callback.


chrisballinger:
@N-Pex Oh, I see, this was a change in XLForm 3.1.2. Should the code currently in our willDisplayCell be moved to cellForRowAtIndexPath as well?

> Update row in cellForRowAtIndexPath instead of willDisplayCell

chrisballinger:
@N-Pex We fixed this in a different way https://github.com/ChatSecure/ChatSecure-iOS/commit/f88c44dc7bb21f1cf7b825245ad1c34e30201945

Issue #550, opened 2016-08-18
Ip addresses in the Usernames are not accepted


sukhovvl:
App shows as incorrect my username which has ip address in it.
For example login@xx.yy.zz.qq instead of login@smth.com.
My username was accepted before I updated the app

chrisballinger:
We havent' seen a lot of demand for this but we can take a look

sukhovvl:
That would be great, thanks

2016-08-18 18:09 GMT+02:00 Chris Ballinger <notifications@github.com>:

ALFriedman:
Any update on this Chris? This is a feature we need as well. Thanks!

afriedmanGlacier:
@chrisballinger This is the same as #539 and I modified my personal version because we needed to login by IP address in the username as well. In OTRXLFormCreator, I modified jidTextFieldRowDescriptorWithValue to use the "OTRUsernameValidator" instead of the "XLFormValidator emailValidatorLong", and it worked fine for us. I am sure you guys had a reason for doing it the other way, so I'm not sure if this is a change you want to use or not.

chrisballinger:
@afriedmanGlacier If you make a PR I'll merge it

Issue #539, opened 2016-07-31
Can't sign in with ip hostname xmpp


joseph-lozano:
I set up a xmpp server for use among my friends. I can't seem to log on ChatSecure though. The sign in page seems to want a TLD, but the domain for the server is just the IP address. I can log in successfully on desktop clients.

Is a TLD on the username/hostname required? Or is there some trick i am missing?

joseph-lozano:
I was able to log in on the android application. It appears to be some kind of client side validation

PR #570, opened 2016-10-07
Allow username with IP address

afriedmanGlacier:
In response to Issues #550 and #539, change Validator in jidTextFieldRowDescriptorWithValue to OTRUsernameValidator instead of emailValidator. An alternate solution might be to get rid of the jidTextFieldRowDescriptorWithValue method or just call the usernameTextFieldRowDescriptorWithValue method from the Username field in Add Existing User


chrisballinger:
Thanks!

PR #665, opened 2017-01-25
Added ability to change avatar

davidchiles:
Haven't been able to test yet. Going to try shortly with zom since the UI to change the avatar is there.


mikechames:
Stop sending these fucking emails I'm fucking tired of it

PR #687, opened 2017-02-10
Move to using optionals instead of force unwrapping

davidchiles:


thundergreen:
When will chatsecure be available in France?

Am 15.02.2017 21:19 schrieb "Chris Ballinger" <notifications@github.com>:

PR #718, opened 2017-03-13
Abort audio recording if no longer in touch

N-Pex:
... which will happen if the "Permission alert" pops up.


chrisballinger:
Thanks!

PR #723, opened 2017-03-22
Post notification when a buddy approves us

N-Pex:


chrisballinger:
@N-Pex Where is `OTRBuddyPendingApprovalDidChangeNotification` handled?

N-Pex:
@chrisballinger Currently only in Zom, https://github.com/zom/Zom-iOS/commit/4a420a8b6fe619eed47af4e53ae685eac8c1eac2 (ZomMainTabbedViewController.swift) but this is maybe functionality that should be available upstream as well?

ChatSecure/Classes/Controllers/XMPP/OTRYapDatabaseRosterStorage.m
```diff
@@ -180,6 +180,11 @@ - (void)updateBuddy:(nullable OTRXMPPBuddy *)buddy withItem:(nonnull NSXMLElemen
     [[OTRDatabaseManager sharedInstance].readWriteDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
         [newBuddy saveWithTransaction:transaction];
     }];
+    
+    if (buddy.pendingApproval && !newBuddy.pendingApproval) {
+        // Buddy was approved
+        [[NSNotificationCenter defaultCenter] postNotificationName:OTRBuddyPendingApprovalDidChangeNotification object:buddy userInfo:nil];
```
chrisballinger: The object parameter should be `self`. If passing the buddy object is required, use a lightweight userInfo dictionary `@{@"buddy": buddy}`.

chrisballinger:
@N-Pex Yeah, I think the rest of it should be upstreamed as well because otherwise this PR doesn't really do anything except add "unused" code. To make the other changes more generic:

1. Add a notification observer on OTRXMPPManager that listens for notifications from the OTRYapDatabaseRosterStorage object, and posts the notifications from there. Eventually this should be moved to a centralized notification observer so the OTRXMPPManager code doesn't rely on UIApplication and can be generic between iOS and macOS.
2. The `String(format: NSLocalizedString("You and %@ are now friends on Zom", comment: "Text for approved buddy notification"), name)` should be added to strings.json and used via the autogenerated OTRStrings.h file. The string should read `You and %@ are now friends.` or similar, and can be overridden in Zom file within the Localizable strings file.

ChatSecure/Classes/Model/Yap Storage/OTRXMPPBuddy.h
```diff
@@ -11,6 +11,8 @@
 
 @class XMPPvCardTemp;
 
+extern NSString * _Nonnull const OTRBuddyPendingApprovalDidChangeNotification;
```
chrisballinger: Add documentation string describing userInfo buddy key.

`/** Contains userInfo with buddy object in "buddy" key */`

ChatSecure/Classes/Controllers/XMPP/OTRYapDatabaseRosterStorage.m
N-Pex: Ok, makes sense, we should probably change the calls in OTREncryptionManager.m as well then, I was kind of inspired by that =)

chrisballinger: @N-Pex I agree, that's some pretty old code over there. I think back then we ignored that the whole point of the "object" was that it acted like a "sender" parameter so you could optimize/filter out notifications from only objects you care about.

N-Pex:
Ok, thanks for having a look, I'll make the changes tomorrow and update the PR!

ChatSecure/Classes/Controllers/XMPP/OTRYapDatabaseRosterStorage.m
N-Pex: @chrisballinger Oh so that's what you "filter" out with setting "object" in the addObserver call, I see. Thanks for pointing that out!

chrisballinger:
@N-Pex While you're at it, shouldn't this PR also send notifications for when you receive a new subscription request?

N-Pex:
@chrisballinger Like this? Not sure if we should have a different message than "WANTS_TO_CHAT()" but using this for now.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
```diff
@@ -836,6 +839,8 @@ -(void)xmppRoster:(XMPPRoster *)sender didReceivePresenceSubscriptionRequest:(XM
         request.accountUniqueId = self.account.uniqueId;
         
         [request saveWithTransaction:transaction];
+        
+        [[UIApplication sharedApplication] showLocalNotificationForSubscriptionRequestFrom:jidStrBare];
```
chrisballinger: @N-Pex Should this only be shown a single time during `if (!request)`?

chrisballinger: Ooops I meant to start a review?

N-Pex: @chrisballinger Yes, that makes sense I think.

chrisballinger:
@N-Pex Looks good! One more thing I just realized, the `showLocalNotificationFor` function should only be called on the main queue, so I'll just dispatch_async all of it internally. It's currently being called from the background which might have funky behavior, especially for the non-iOS 10 code path. I'll merge this as-is and fix that on my end.

N-Pex:
@chrisballinger Great, thanks!

PR #739, opened 2017-04-18
Dukgo migration

N-Pex:
@chrisballinger Probably should not be merged to master, but some new feature branch?

This PR adds the "migration info" as a table view header for the chats view and currently pops up the "createNewAccount" view controller from onboarding when "Get Started" is tapped.

### DIFF
--- a/ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.h
+++ b/ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.h
@@ -71,6 +71,10 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setAvatar:(UIImage *)newImage completion:(void (^)(BOOL success))completion;
 
+/** Force a vCard update (by manipulating pixel values in the avatar image)
+ */
+- (void)forcevCardUpdateWithCompletion:(void (^)(BOOL success))completion;
+
 - (void)changePassword:(NSString *)newPassword completion:(void (^)(BOOL,NSError*))completion;
 
 /** Will try to send a probe to fetch last seen. If buddy is still pendingApproval it will retry subscription request. */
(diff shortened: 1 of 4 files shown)


chrisballinger:
Awesome, thank you! I'll merge this into a feature branch and continue where you left off.

N-Pex:
@chrisballinger Great! I just added a commit for the "It looks like your buddy has migrated" view we talked about (in the messages view controller). The text (and probably design) needs to be changed, but this at least adds the basic functionality.

N-Pex:
@chrisballinger Ok, I think I implemented most of the changes we talked about now.

chrisballinger:
Merged your changes into `dukgo_migration` branch and fixed detection https://github.com/ChatSecure/ChatSecure-iOS/commit/8f04dbf942ae284f7d40974ccafaa0a89f230b36. You can pull in my branch to yours and then if you add more commits to your branch, I'll change the PR base from master to my branch because GitHub won't let me change it until there are more commits in your remote branch.

N-Pex:
@chrisballinger Ok, how about now?

N-Pex:
I added a method to force the vCard update (by slightly modifying the avatar image). I guess you should call this after setting the vCard.jid in the migration controller =)

chrisballinger:
@N-Pex Awesome, thanks. Can you explain a bit about what the avatar modification code is doing and where it's from? I think you could make it a lot shorter by simply calling `setAvatar:completion:` with the existing `avatarImage` UIImage. Because it recompresses the image with JPEG anyway, it will be a different hash.

```
[self setAvatar:img completion:completion];
```

N-Pex:
@chrisballinger Oh, I didn't realize it's compressed. No way to do it cleanly then, so maybe this commit is pretty useless. Anyway, it's based on pixel manipulation code I wrote for a previous project. It sets the bottom rightmost pixel to the same value as the pixel beside it, with the blue component being +- 1 from the current value (just to force a new hash). But, since we'll get a new hash anyway...

chrisballinger:
@N-Pex Ya fortunately the compression level results in a negligible quality loss from recompression, especially at the small size in the avatar.

PR #755, opened 2017-04-26
Allow the account detail view controller to be overridden

N-Pex:
This is needed by Zom to implement the "set as default" account option. Other ways to accomplish this are welcome.


chrisballinger:
Looks good to me!

PR #759, opened 2017-05-04
Migration notification

N-Pex:
Not sure if this is functionality you want upstream or not?


ChatSecure/Classes/Categories/UIApplication+ChatSecure.swift
```diff
@@ -79,17 +79,41 @@ public extension UIApplication {
...
+                    localNotification.threadIdentifier = identifier
                 }
-                let request = UNNotificationRequest(identifier: UUID().uuidString, content: localNotification, trigger: nil) // Schedule the notification.
+                if let userInfo = userInfo {
+                    localNotification.userInfo = userInfo
+                }
+                var trigger:UNNotificationTrigger? = nil
+                if recurring {
+                    var date = DateComponents()
+                    date.hour = 11
```
chrisballinger: @N-Pex Does this mean it repeats every 11 hours?

ChatSecure/Classes/OTRAppDelegate.m
```diff
@@ -569,7 +569,9 @@ - (void) userNotificationCenter:(UNUserNotificationCenter *)center willPresentNo
 
 - (void) userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNotificationResponse:(UNNotificationResponse *)response withCompletionHandler:(void (^)())completionHandler {
     NSDictionary *userInfo = response.notification.request.content.userInfo;
-    if (userInfo[kOTRNotificationThreadKey] == nil) {
+    if ([kOTRNotificationTypeNone isEqualToString:userInfo[kOTRNotificationType]]) {
+        // Nothing
+    } else if ([kOTRNotificationTypeSubscriptionRequest isEqualToString:userInfo[kOTRNotificationType]]) {
```
chrisballinger: @N-Pex I think this should be done in reverse because userInfo[kOTRNotificationType] can be nil.

```obj-c
NSString *type = userInfo[kOTRNotificationType];
if (![type isKindOfClass:[NSString class]]) { return; } // handles nil or wrong class
if ([type isEqualToString:.....]) etc
```

ChatSecure/Classes/View Controllers/OTRConversationViewController.m
```diff
@@ -135,6 +135,33 @@ - (void) showOnboardingIfNeeded {
...
         self.hasPresentedOnboarding = YES;
     }
+    
+    OTRXMPPAccount *needsMigration = [self checkIfNeedsMigration];
+    if (needsMigration != nil) {
+        // Show local notification prompt
+        OTRServerDeprecation *deprecationInfo = [OTRServerDeprecation deprecationInfoWithServer:needsMigration.bareJID.domain];
+        if (deprecationInfo != nil) {
+            NSString *notificationBody;
+            NSDate *now = [[NSDate alloc] initWithTimeIntervalSinceNow:0];
```
chrisballinger: @N-Pex You can also say `[NSDate date]`

ChatSecure/Classes/Categories/UIApplication+ChatSecure.swift
chrisballinger: @N-Pex Does this mean it repeats every 11 hours?

ChatSecure/Classes/View Controllers/OTRConversationViewController.m
```diff
@@ -135,6 +135,33 @@ - (void) showOnboardingIfNeeded {
         }
         self.hasPresentedOnboarding = YES;
     }
+    
+    OTRXMPPAccount *needsMigration = [self checkIfNeedsMigration];
+    if (needsMigration != nil) {
+        // Show local notification prompt
+        OTRServerDeprecation *deprecationInfo = [OTRServerDeprecation deprecationInfoWithServer:needsMigration.bareJID.domain];
+        if (deprecationInfo != nil) {
+            NSString *notificationBody;
```
chrisballinger: Init with a value, so you don't need else case:

```objc
NSString *notificationBody = [NSString stringWithFormat:MIGRATION_NOTIFICATION_STRING(), deprecationInfo.name];
```

ChatSecure/Classes/View Controllers/OTRConversationViewController.m
```diff
@@ -135,6 +135,33 @@ - (void) showOnboardingIfNeeded {
...
+                NSCalendar *gregorianCalendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
+                NSDateComponents *components = [gregorianCalendar components:NSCalendarUnitDay
+                                                                    fromDate:now
+                                                                      toDate:deprecationInfo.shutdownDate
+                                                                     options:0];
+                long days = [components day];
+                notificationBody = [NSString stringWithFormat:MIGRATION_NOTIFICATION_WITH_DATE_STRING(), days];
+            } else {
+                // No shutdown date or already passed
+                notificationBody = [NSString stringWithFormat:MIGRATION_NOTIFICATION_STRING(), deprecationInfo.name];
```
chrisballinger: See above comment

OTRAssets/Strings/OTRStrings.h
```diff
@@ -10,6 +10,8 @@ FOUNDATION_EXPORT NSString* ABOUT_VERSION_STRING();
 FOUNDATION_EXPORT NSString* ACCOUNTS_STRING();
 /** "Account", The string describing a buddy's account */
 FOUNDATION_EXPORT NSString* ACCOUNT_STRING();
+/** "Active", segmented control value for active friends */
+FOUNDATION_EXPORT NSString* ACTIVE_BUDDIES_STRING();
```
chrisballinger: We already have a string for "Active" called `ACTIVE_STRING()`, should we just use that? Perhaps this would be translated differently in other languages. One is implying "Last user interaction" and the other is "Not archived".

ChatSecure/Classes/Categories/UIApplication+ChatSecure.swift
N-Pex: @chrisballinger No, it means that it repeats every day at 11AM.

OTRAssets/Strings/OTRStrings.h
N-Pex: @chrisballinger Yes, I wondered about that, but decided against reuse because the comment says something about "Will be shown in the UI as Active x ago".

ChatSecure/Classes/Categories/UIApplication+ChatSecure.swift
chrisballinger: Oh yeah my bad. So this will spam the user every morning until they migrate?

ChatSecure/Classes/OTRAppDelegate.m
N-Pex: No, I believe this will work as intended. If userInfo[kOTRNotificationType] is nil the isEqualToString will return NO and we'll fall through to the final else-clause.

ChatSecure/Classes/View Controllers/OTRConversationViewController.m
N-Pex: Ok, sounds good.

ChatSecure/Classes/View Controllers/OTRConversationViewController.m
N-Pex: Thanks, looks a lot nicer =)

ChatSecure/Classes/OTRAppDelegate.m
chrisballinger: From the Apple docs for `NSString.h`:

```objc
NS_ASSUME_NONNULL_BEGIN
...
- (BOOL)isEqualToString:(NSString *)aString;
```

The method assumes `aString` is nonnull, so even if it doesn't crash, it's not what the API wants. For instance, in Swift it wouldn't even compile. It would be better to at least swap the order so you're sending a message to nil, instead of passing a nil parameter.

N-Pex: Ok, fair enough, change submitted! Just my background in other languages showing through - "sending a message to nil" is to me a pretty outrageous operation... =)

chrisballinger: Yeah Obj-C is kind of weird, but allows for "clean looking" code by chaining nil messages. In Swift you have to `explicitly?.use?.optional?.chaining` which is probably better.

ChatSecure/Classes/Categories/UIApplication+ChatSecure.swift
N-Pex: Yes, exactly

PR #761, opened 2017-05-08
Don't send yourself a message during migration

N-Pex:


ChatSecure/Classes/View Controllers/Login View Controllers/OTRAccountMigrationViewController.m
```diff
@@ -109,6 +109,10 @@ - (void) handleSuccessWithNewAccount:(OTRXMPPAccount*)newAccount sender:(id)send
         // If spamming friends, create some messages for them
         if (shouldSpamFriends) {
             [buddies enumerateObjectsUsingBlock:^(OTRXMPPBuddy * _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+                // Don't send yourself a message
+                if ([obj.bareJID isEqualToJID:self.oldAccount.bareJID options:XMPPJIDCompareBare]) {
+                    return;
+                }
```
chrisballinger: @N-Pex Looks like this could be cleaned up by moving the `shouldSpamFriends` logic into the buddy enumeration block below.

chrisballinger:
Fixed it in master

PR #762, opened 2017-05-08
Don't show presence requests in archive view

N-Pex:


N-Pex:
Not sure about the second commit here, because that would require the user to "unarchive" an archived thread just to read it. But maybe that's desired.

N-Pex:
Ok, removed the filtering of notifications for archived threads. Unfortunately this here will lead to different branches between CS and Zom, I don't see any straightforward way to override this in Zom, you can't extend an extension.

chrisballinger:
@N-Pex Thanks. We can move these from a class extension to an overridable class via the theme, but it would be unwieldy I think.

I think the ability to 'mute' a thread is useful, but orthogonal to archiving, which is more to tidy up your list. I think a better long to solution is to add an `muteExpiration` date property to the `OTRThreadOwner` protocol, and silence notifications from muted threads. We don't have time to expose this in the UI, but Zom could set that to `[NSDate distantFuture]` during the migration. Then later down the line we could add the ability to selectively mute/unmute a thread. Signal exposes this option in their buddy detail view, and allows you to mute a thread for the following durations: 1 hour, 1 day, 1 week, 1 year.

chrisballinger:
@N-Pex I have added a readwrite `muteExpiration` date and `isMuted` convenience accessor in this commit to master: 2eaf004e

chrisballinger:
You'll still need to add a check for `isMuted` in the notification code and submit a PR for that, but you should now be able to add `muteExpiration = [NSDate distantFuture]` in Zom's post-migration step.

PR #766, opened 2017-05-10
Add migration type link and send it in migration message

N-Pex:
@chrisballinger I could not easily append m=1, changing to /m/ was easier (because how the url is encoded/decoded). Let me know if this will work?

### DIFF
--- a/ChatSecure/Classes/Categories/NSURL+ChatSecure.m
+++ b/ChatSecure/Classes/Categories/NSURL+ChatSecure.m
@@ -110,8 +110,11 @@
     NSArray *components = [urlString componentsSeparatedByString:@"/i/#"];
     
     if (components.count != 2) {
-        completion(nil, nil);
-        return;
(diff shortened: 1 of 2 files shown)


chrisballinger:
I took an alternative approach, which you'll see in the master branch. I couldn't figure out how to intercept the https:// links to show the add-buddy dialog directly, but I made sure the XMPP URIs on the resulting page worked better. While I was at it, I made it so you can't add yourself to yourself, and there's an `isArchived` computed property on OTRAccount that checks if there's a forwarding JID set. That way when you are handling an invite, there's a smaller list of accounts.

chrisballinger:
There seems to be no way to properly ingest the migration bit `m=1`, because the link just opens Safari, and then from there you just have a regular old XMPP URI and lose the context of who is the "old" buddy. There are probably ways around this but I think it will take longer than the current deadline.

PR #783, opened 2017-06-13
Group chat fixes

N-Pex:


ChatSecure/Classes/View Controllers/OTRMessagesGroupViewController.m
```diff
@@ -44,6 +44,11 @@ - (void)setupGroupChat:(NSArray <NSString *>*)buddies account:(OTRAccount *)acco
     [self setThreadKey:self.threadKey collection:[OTRXMPPRoom collection]];
 }
 
+- (nullable OTRAccount *)accountWithTransaction:(nonnull YapDatabaseReadTransaction *)transaction {
```
chrisballinger: @N-Pex I think this already exists in the superclass and you could just expose the method in a private header / class extension?

N-Pex: @chrisballinger Yes, it exists but returns nil.

Good thing I wrote a comment in the commit =)

"super will just return nil, since the ThreadKey has not been set when
this is called during setup."

chrisballinger: Ahhhh ok, I missed that because GitHub does a good job of hiding the extended commit messages. Perhaps add an inline comment that this is overriding super's behavior bc of that?

Btw it looks like your computer's date is wrong, says these commits are from April 12?

N-Pex: Ok, added a comment.

Hahaha, sorry, no, they ARE that old. I started looking at some group chat issues back then, but then the whole migration thingy came along and the changes never made it to a pull request...

chrisballinger: Hahahaha thanks!

PR #784, opened 2017-06-15
Run migration message detection on non-UI thread.

N-Pex:
Also, run it once for each message, not every time it scrolls by.

@chrisballinger What do you think of this solution?

### DIFF
--- a/ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
+++ b/ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
@@ -1907,9 +1907,9 @@ heightForCellBottomLabelAtIndexPath:(NSIndexPath *)indexPath
         __typeof__(self) strongSelf = weakSelf;
         __block bool foundLink = false;
 
-        for (int i = range.location; !foundLink && i < (range.location + range.length); i++) {
+        for (NSUInteger i = range.location; !foundLink && i < (range.location + range.length); i++) {
             id <OTRMessageProtocol>message = [strongSelf messageAtIndexPath:[NSIndexPath indexPathForRow:i inSection:0]];
-            if (message != nil && [message messageIncoming]) {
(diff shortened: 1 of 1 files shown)


ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -1945,4 +1929,37 @@ - (IBAction)didPressMigratedSwitch {
     }
 }
 
+// If we find any incoming migration link messages, show the "your friend has migrated" header
+// to allow the user to start chatting with the new account instead.
+- (void) checkRangeForMigrationMessage:(NSRange)range {
```
chrisballinger: This can still cause stuttering because of the database fetch `messageAtIndexPath:` operates serially. I'd suggest passing in a messageText string parameter. Also, I don't think the ivar for linkDetector is needed, you could just allocate a temporary one in here, and I'm not sure if its thread safe to reuse it anyway.

Is this possible to move this change downstream as well into ZomMessagesViewController?

N-Pex: Well, then we might have another problem, because messageAtIndexPath is called ALL THE TIME by the other code, right?

Anyway, I moved the coded downstream instead. I still think the delegate is needed though (the last change still present here) for in-app link handling. But I guess I can move that into downstream as well.

chrisballinger: It's ok to call messageAtIndexPath within the methods that are dequeueing cells and whatnot from the main thread, but the issue was that the global dispatch queue will be calling it from all sorts of background threads, so at times messageAtIndexPath will block the other threads (including the main thread). 

The in-app link handling and delegate code is fine. 👍

N-Pex:
Closing this, adding new with the minimal change set.

PR #787, opened 2017-06-17
Do not close stream when in background

stigger:
Keeping the stream alive when going to background was broken in commit 158bb562d4e32f241674bbb453043be939e08c6b, where `[xmppStream disconnect]` was replaced with `disconnectAfterSending`. These two methods have different semantics: `disconnectAfterSending` sends `</stream>` before closing the socket, and `disconnect` does not.

This solution is not very elegant, but works and does not require changing XMPPFramework.


chrisballinger:
Ahh thank you! This issue was being reported by quite a few people but I couldn't figure out how it was possible.

PR #791, opened 2017-06-21
Fix typo and allow deliveryString to be attributed

N-Pex:
That way overrides can use images instead of just text.


N-Pex:
By the way, line 1549 in OTRMessagesViewController.m seems to have some thought behind it, but the attribute is never used. Dead code? (in that case we can remove lines 1531 and 1532 as well).

chrisballinger:
Build is failing?

N-Pex:
Is it? Travis shows green here(?)

PR #801, opened 2017-06-27
Revert to old isPendingApproval implementation

N-Pex:
This reverts the change that breaks the two-way subscription handling.

Without having looked at it too deeply I think it really strange that this implementation differs logically from those in XMPPUserCoreDataStorageObject.m and XMPPUserMemoryStorageObject.m. Why should they not be the same?


N-Pex:
Ok, I see now that they are not actually directly related, they just have the same name, so there could actually be subtle differences in use case. In that case different logic can make sense of course. My bad. (Original code came from here: https://github.com/ChatSecure/ChatSecure-iOS/commit/3c9f26aeac2ab0e91a8b9e86f68046e6e5211b85)

chrisballinger:
Thanks for tracking that down!

PR #802, opened 2017-06-28
Group chat fixes

N-Pex:
Fix crash on receiving group chat message and setup room correctly when creating.


chrisballinger:
Thx!

PR #809, opened 2017-07-05
Friending problems

N-Pex:
I think this fixes most of the one-way subscription and OMEMO-not-starting problems.


chrisballinger:
Looks good, thank you!

Issue #470, opened 2016-04-15
UI slowdown with large message history


chrisballinger:
We need to paginate JSQ message history to avoid slowing down the UI when people have long message history

vtrullinger:
Anything you could do to speed up the UI when there's a long message history would be fantastic!

chrisballinger:
@vtrullinger Are you getting extremely long pauses that initially appear to be deadlocks?

vtrullinger:
Yes, that's exactly it!

On Jan 29, 2017, at 6:18 PM, Chris Ballinger <notifications@github.com<mailto:notifications@github.com>> wrote:


@vtrullinger<https://github.com/vtrullinger> Are you getting extremely long pauses that initially appear to be deadlocks?

-

Issue #360, opened 2015-07-10
app Freeze


JackQiangXJ:
when in message screen, keep receive message from buddy, app will freeze 1s to 2s.

I think it caused by  [self.collectionview reloaddata] in the func ----- yapDatabaseModified:

chrisballinger:
Does this happen all the time? Do you have a large number of messages? It
may be an issue with a 3rd party framework.

JackQiangXJ:
yes,it happen all the time,I think you can receive 50 offline messages onetime.in my opinion,it caused by when received message,app will get a yapdbnotification in messageviewcontroller, then you will reload all data in this screen.

JackQiangXJ:
yes,it happen all the time,I think you can try to receive 50 offline messages onetime.in my opinion,it caused by when received message,app will get a yapdbnotification in messageviewcontroller, then you will reload all data in this screen.

chuquangbao:
Yeah, I have this same problems. 
And, I have more 2000 messages. App free when i am typing ( send chat and save chat state).
How can i solve it?

chrisballinger:
We should limit the number of messages to some same default in the Yap
mappings, and provide a way to scroll back through your history and load
that dynamically.

PR #812, opened 2017-07-08
Chatview performance

stigger:
The current version does not work very well with large (thousands of messages) histories. The single fact that ChatSecure loads all messages at once affects many areas:
- initial opening;
- loading new messages due to a push notification;
- overall responsiveness (scrolling, typing).

Without going into details, the issue is that too much data is being processed to often without any reason to do so (and often on the UI thread).

There are two main changes in this pull request:
- instead of loading the entire history, load only the last 50 messages, and load the rest on demand;
- handle message loading more efficiently by caching and offloading to background threads the most heavy calculations.

Closes at least #360 & #470, maybe some others.


chrisballinger:
Awesome thank you! I'm away at the moment but I'll plan to review this next week

chrisballinger:
I really like what you've done here! Works really well! Apologies that it took me so long to review, I was away the last couple of weeks, but am now back working full time.

stigger:
Thanks for merging! I'd really like to see a TestFlight build with these changes: ChatSecure became practically unusable for me because of a large history.

chrisballinger:
Definitely. I found some issues w/ the PR after the merge, so I'm fixing that up.

The optimization to pre-calclulate the bubble size makes the Xcode 9 UI thread checker complain, so I've removed that part for now.

stigger:
I'd say that it's safe to ignore the UI thread checker. Accessing UIKit on background threads is not inherently a bad thing. In this case we're not modifying anything, only accessing already existing structures.

In the worst case (if the `JSQMessagesCollectionViewFlowLayout` changes on the UI thread while we're calculating) we might end up with incorrect bubble size, but I don't see how that's possible. Still, I'll try to submit a PR to JSQMessagesViewController that allows to implement this without accessing UIKit on background..

chrisballinger:
Too bad JSQ was abruptly shut down and no community fork has emerged... :( https://www.jessesquires.com/blog/officially-deprecating-jsqmessagesviewcontroller/

Yeah I don't disagree with you about the verbosity of the main thread checker, but I err on the side of caution because I can't see the internals of UIKit. Even without that block though, your PR still speeds things up dramatically!

michaelkirk:
Hi, we recently fixed some similar looking crashes in Signal. 

This may or may not be useful: https://github.com/WhisperSystems/JSQMessagesViewController/pull/14/files

Issue #817, opened 2017-07-27
Crash when scrolling quickly through chat history


chrisballinger:
Regression caused by #812

cc @stigger 

```
Application Specific Information:
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'layout attributes for supplementary item at index path (<NSIndexPath: 0xc000000000000016> {length = 2, path = 0 - 0}) changed from <JSQMessagesCollectionViewLayoutAttributes: 0x13398d420> index path: (<NSIndexPath: 0xc000000000000016> {length = 2, path = 0 - 0}); element kind: (UICollectionElementKindSectionHeader); frame = (0 0; 414 32); zIndex = 10;  to <JSQMessagesCollectionViewLayoutAttributes: 0x131ef8640> index path: (<NSIndexPath: 0xc000000000000016> {length = 2, path = 0 - 0}); element kind: (UICollectionElementKindSectionHeader); frame = (0 0; 414 32); zIndex = -1;  without invalidating the layout'

Thread 0 Crashed:
0   CoreFoundation                  0x000000018878afe0 __exceptionPreprocess + 124
1   libobjc.A.dylib                 0x00000001871ec538 objc_exception_throw + 52
2   CoreFoundation                  0x000000018878aeb4 +[NSException raise:format:arguments:] + 100
3   Foundation                      0x0000000189223760 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 108
4   UIKit                           0x000000018e9195c8 __45-[UICollectionViewData validateLayoutInRect:]_block_invoke + 1324
5   UIKit                           0x000000018e918ad4 -[UICollectionViewData validateLayoutInRect:] + 1492
6   UIKit                           0x000000018f1fb184 -[UICollectionViewData layoutAttributesForCellsInRect:validateLayout:] + 144
7   UIKit                           0x000000018f1b79f4 -[UICollectionView _computePrefetchCandidatesForVisibleBounds:futureVisibleBounds:prefetchVector:notifyDelegateIfNeeded:] + 128
8   UIKit                           0x000000018f1b7954 -[UICollectionView _computePrefetchCandidatesForVelocity:notifyDelegateIfNeeded:] + 164
9   UIKit                           0x000000018f1b7884 -[UICollectionView _prefetchItemsForVelocity:maxItemsToPrefetch:invalidateCandidatesOnDirectionChanges:] + 764
[...]

PR #823, opened 2017-07-29
Fixing a crash and a couple of other issues introduced in a recent OTRMessagesViewController rework (#817)

stigger:


chrisballinger:
Does this also fix the occasional issue when receiving messages where the incoming message flashes and then disappears (or does not appear at all until manually refreshing the view)? I'm not sure if it's related, but I hadn't seen it before merging the PR.

stigger:
Yes, I believe it should be fixed as well.

chrisballinger:
Thanks!

PR #831, opened 2017-08-04
Group profiles

N-Pex:


ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
+        self.headerRows.append("addFriends")
+        self.headerRows.append("mute")
+        self.headerRows.append("members")
+        self.footerRows.append("leave")
+        
+        let image = OTRGroupAvatarGenerator.avatarImage(withUniqueIdentifier: self.roomKey!, width: Int(largeAvatarView.frame.width), height: Int(largeAvatarView.frame.height))
+        largeAvatarView.image = image
+
+        //Setup Table View
+        if (self.tableView == nil) {
```
chrisballinger: If you're using storyboards setting up the table view should be unnecessary.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
-        self.tableView.autoPinEdgesToSuperviewEdges()
+    override open func viewDidLoad() {
+        super.viewDidLoad()
+        
+        self.headerRows.append("name")
+        self.headerRows.append("share")
+        self.headerRows.append("addFriends")
+        self.headerRows.append("mute")
+        self.headerRows.append("members")
+        self.footerRows.append("leave")
```
chrisballinger: Please use an enum here instead of strings. For example: https://github.com/ChatSecure/ChatSecure-iOS/blob/master/ChatSecure/Classes/View%20Controllers/AccountDetailViewController.swift#L24

```swift
enum TableSections: Int {
    case account
    case invite
    case details
    case loginlogout
    case migrate
    case delete
    static let allValues = [account, invite, details, loginlogout, migrate, delete]
}
```

This way you can enumerate TableSections.allValues, and when you add a section, you need to handle it in your switch/case (don't implement `default`).

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
-    let viewHandler: OTRYapViewHandler
-    let roomKey: String
+    @IBOutlet open weak var tableView:UITableView!
+    @IBOutlet weak var largeAvatarView:UIImageView!
+    
+    open var viewHandler:OTRYapViewHandler?
+    open var roomKey:String?
+    open var room:OTRXMPPRoom?
+    open var headerRows:[String] = []
+    open var footerRows:[String] = []
```
chrisballinger: Are these marked as `open` for a reason, or can some of them be `private`?

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
 
-    private func setupViewHandler() {
-        viewHandler.delegate = self
-        viewHandler.setup(DatabaseExtensionName.groupOccupantsViewName.name(), groups: [roomKey])
+    public func setupViewHandler(databaseConnection:YapDatabaseConnection, roomKey:String) {
+        self.roomKey = roomKey
+        databaseConnection.read({ (transaction) in
+            if let key = self.roomKey {
+                self.room = OTRXMPPRoom.fetchObject(withUniqueID: key, transaction: transaction)
+            }
```
chrisballinger: Is it necessary to hold on to references for both `room` and `roomKey`? Why not just `room`?

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
+                    if let room = OTRXMPPRoom.fetchObject(withUniqueID: key, transaction: transaction) {
+                        cell.textLabel?.text = room.subject
+                        cell.detailTextLabel?.text = "" // Do we have creation date?
+                    }
+                })
+            }
+            cell.selectionStyle = .none
+            return cell
+        case "share":
+            let cell = tableView.dequeueReusableCell(withIdentifier: "cellGroupShare", for: indexPath)
```
chrisballinger: Make an enum for cell identifiers instead of using magic strings. You could even create a mapping between enums.

```swift
enum CellType {
case group
var cellIdentifier: CellIdentifier {
return .groupShare
}
}
enum CellIdentifier: String {
case groupShare
}
```

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
+            
+        default:
+            return UITableViewCell()
+        }
+    }
+    
+    open func createFooterCell(indexPath:IndexPath, type:String) -> UITableViewCell {
+        switch type {
+        case "leave":
+            let cell = tableView.dequeueReusableCell(withIdentifier: "cellGroupLeave", for: indexPath)
```
chrisballinger: Use enum here as well

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -10,89 +10,221 @@ import Foundation
...
+            return createHeaderCell(indexPath: indexPath, type: headerRows[indexPath.row])
+        } else if isFooterSection(section: indexPath.section) {
+            return createFooterCell(indexPath: indexPath, type: footerRows[indexPath.row])
+        }
+        let adjustedIndexPath = IndexPath(row: indexPath.row, section: indexPath.section - 1)
+        let cell:OTRBuddyInfoCell = tableView.dequeueReusableCell(withIdentifier: OTRRoomOccupantsViewController.CellIdentifier, for: indexPath) as! OTRBuddyInfoCell
+        var buddy:OTRXMPPBuddy? = nil
+        if let roomOccupant = self.viewHandler?.object(adjustedIndexPath) as? OTRXMPPRoomOccupant, let room = self.room {
+            OTRDatabaseManager.shared.readOnlyDatabaseConnection?.read({ (transaction) in
+                    buddy = OTRXMPPBuddy.fetch(withUsername: roomOccupant.realJID ?? roomOccupant.jid!, withAccountUniqueId: room.accountUniqueId!, transaction: transaction)
```
chrisballinger: Can you avoid force unwrap?

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Yes, subclasses should be able to access headers and footers to add/remove/config at will.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Ok, let's just use room then, I guess I was lazy not removing all old refs to roomKey when I added room =)

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Yes, I know a lot of the code uses this approach, but this causes trouble when subclasses want to do other stuff, like add a row for "send my dog flowers" or whatever, since the enums can't be extended. That's why I went for a strings-based approach.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Yes, I left this for backward compatibility if someone else is using the code, but I am happy to remove it.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: In my first approach I had a mapping string->cellIdentifier but since we have to do some configuration of the cells anyway (currently only for the "group name" cell) I felt that the mapping was pretty unnecessary and that the cellForRow... could be responsible for fetching the right type of cell.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Ah, yes, I'll fix!

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
chrisballinger: I see. In that case, at the very least please use static string constants instead of raw strings. There is likely a better approach with an array containing some simple structs, but I'll have to think about it a bit more.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
chrisballinger: In the earlier version of the code I experienced an unexpected nil crash related to self.tableView being nil even after the init/assignment, which I thought was kind of strange. If this view isn't used/tested outside of the context of a storyboard, I think think the manual path can be left out.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
chrisballinger: Sure, but please use static string constants for your cell identifiers instead of raw strings.

N-Pex:
Used static strings now and instead of magic names let them be cell identifiers.

Moved fonts to OTRResources folder (you should consider moving FontAwesome as well)

chrisballinger:
Thanks for the changes. After testing it a bit, I think it would be more appropriate to merge into an upstream feature branch `group_profiles` until it is more complete. Although it's an improvement over the status quo, it's too unstable to merge into master at the moment.

chrisballinger:
@N-Pex Please fix:

1. Square avatars
2. "Members" section title is a cell and selectable? If using a table view, ideally that should be a section header not a cell. Same with the room title.
3. Uneven icon spacing on horizontal axis: Spacing for "share" icon doesn't match the other icons. "Share" should be shifted to the right, and "add friends" should possibly be shifted left slightly.

chrisballinger:
Merged into #842

PR #842, opened 2017-08-29
Unify chat controllers

N-Pex:
Maybe this pull request should be opened against another branch...

Anyway, this is really just the last commit here (the other changes are in the other PR). I tried to move over all functionality of group chat to the "normal" message view controller, to unify things. Seems to work ok here, but obviously needs testing! Let's discuss. Just wanted to create this PR here so we can avoid us both doing the same changes =)


ChatSecure/Classes/View Controllers/OTRMessagesViewController.h
```diff
@@ -52,6 +54,12 @@
 - (nullable OTRXMPPManager *)xmppManagerWithTransaction:(nonnull YapDatabaseReadTransaction *)transaction;
 - (nullable id <OTRMessageProtocol,JSQMessageData>)messageAtIndexPath:(nonnull NSIndexPath *)indexPath;
 
+- (BOOL)canSendMedia;
+
+/** Group chat support */
```
chrisballinger: Perhaps these could be moved to a category.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.h
```diff
@@ -64,4 +72,7 @@
 /** override this method to customize what should be shown at the beginning of the message status */
 - (nullable NSAttributedString *) encryptionStatusStringForMessage:(nonnull id<OTRMessageProtocol>)message;
 
+/** Override this to return a fallback conference service JID. Will only be called if no conference server was discovered. Return nil to abort group chat creation. May be used to show warning message to user. */
+- (nullable NSString *)getFallbackConferenceServiceJID;
```
chrisballinger: This shouldn't really be in here. I'd prefer not to have any reference to this hack upstream at all. I really don't agree with using a client hack to work around a trivial server configuration issue.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -1227,6 +1253,10 @@ - (void)didFinishTyping {
 
 #pragma - mark Sending Media Items
 
+- (BOOL)canSendMedia {
+    return self.state.canSendMedia || [self isGroupChat];
```
chrisballinger: This isn't entirely accurate. I think it should just be `self.state.canSendMedia`, because otherwise you will think you can send media in group chats even if you can't.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.h
```diff
@@ -52,6 +54,12 @@
 - (nullable OTRXMPPManager *)xmppManagerWithTransaction:(nonnull YapDatabaseReadTransaction *)transaction;
 - (nullable id <OTRMessageProtocol,JSQMessageData>)messageAtIndexPath:(nonnull NSIndexPath *)indexPath;
 
+- (BOOL)canSendMedia;
```
chrisballinger: This should be removed because the state should be reflected accurately. It shouldn't return true just because it's a group chat.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -2123,4 +2196,29 @@ - (IBAction)didPressMigratedSwitch {
...
+    }
+    if (service != nil) {
+        NSString *roomName = [NSUUID UUID].UUIDString;
+        XMPPJID *roomJID = [XMPPJID jidWithString:[NSString stringWithFormat:@"%@@%@",roomName,service]];
+        self.threadKey = [xmppManager.roomManager startGroupChatWithBuddies:buddies roomJID:roomJID nickname:account.username subject:name];
+        [self setThreadKey:self.threadKey collection:[OTRXMPPRoom collection]];
+    }
+}
+
+- (NSString *)getFallbackConferenceServiceJID {
```
chrisballinger: Nathan needs to fix this on the server.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -123,38 +154,40 @@ open class OTRRoomOccupantsViewController: UIViewController {
...
+        self.tableHeaderView?.setView(OTRRoomOccupantsViewController.HeaderCellMute, hidden: muted)
+        self.tableHeaderView?.setView(OTRRoomOccupantsViewController.HeaderCellUnmute, hidden: !muted)
+    }
+    
+    open func createHeaderCell(type:String) -> UITableViewCell {
+        var cell:UITableViewCell?
         switch type {
         case OTRRoomOccupantsViewController.HeaderCellGroupName:
-            let cell = tableView.dequeueReusableCell(withIdentifier: type, for: indexPath)
+            cell = tableView.dequeueReusableCell(withIdentifier: type)
```
chrisballinger: Why no longer using the `indexPath` version? If it's crashing you just need to register the cell/nib for the identifier.

chrisballinger:
@N-Pex I've also noticed a bug when switching between group and 1:1 chats that the navbar title and state doesn't get updated properly, so it looks like you can send OMEMO messages to a group.

Btw, feel free to pull my upstream `unify_chat_controllers` branch for some changes I've made toward supporting group media messaging.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
N-Pex: Well, we are just using the fact that the tableview owns the designs for the cells and knows how to instantiate them. We are not actually _using_ the cells in the table, we add their views to the header stack view. That's why we don't use the IndexPath. I guess another option would be to create xibs for the different cell types, I just find it convenient to have them all gathered in the tableView in the storyboard.

N-Pex:
Ok, changes made, need to look at the final issue with navbar/state tomorrow.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomManager.m
```diff
@@ -133,6 +133,27 @@ - (void)inviteUser:(XMPPJID *)user toRoom:(XMPPJID *)roomJID withMessage:(NSStri
     [room inviteUser:user withMessage:message];
 }
 
+- (void)inviteBuddies:(NSArray<NSString *> *)buddyUniqueIds toRoom:(XMPPRoom *)room {
+    if (!buddyUniqueIds.count) {
+        return;
+    }
+    NSMutableArray<OTRXMPPBuddy*> *buddies = [NSMutableArray arrayWithCapacity:buddyUniqueIds.count];
+    [self.inviteDictionary removeObjectForKey:room.roomJID.bare];
```
chrisballinger: What is the `inviteDictionary` doing here exactly? There is a risk of mutating it from multiple threads. I think it should be moved back to where it was extracted below.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomManager.m
```diff
@@ -308,23 +329,7 @@ - (void)xmppRoomDidJoin:(XMPPRoom *)sender
         
         //Invite buddies
         NSArray<NSString*> *buddyUniqueIds = [self.inviteDictionary objectForKey:sender.roomJID.bare];
-        if (!buddyUniqueIds.count) {
```
chrisballinger: this probably belongs here
```objc
[self.inviteDictionary removeObjectForKey:room.roomJID.bare];
```

N-Pex: Oops, you're right, that should have remained where it was.

N-Pex:
@chrisballinger Could you have an extra close look at the last commit here, "leave" please? I saw a lot of crashes because JSQMessageController was accessing data that was already deleted if I just deleted the room and then "popped" controllers back the root, so I had to implement the part with allowing "nil" value in setThreadKey. Can you think of a better way to do this?

chrisballinger:
What specifically is the crash?

In theory it should already allow nil because on iPad it will startup with a nil conversation in split view.

N-Pex:
Yes, but only on startup, it did not allow resetting it back to nil (the viewHandler would still be the old one, complete with old mappings...)

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -12,8 +12,14 @@ import PureLayout
...
 import OTRAssets
 
+@objc public protocol OTRRoomOccupantsViewControllerDelegate {
+    func didLeaveRoom(_ roomOccupantsViewController: OTRRoomOccupantsViewController) -> Void
+}
+
 open class OTRRoomOccupantsViewController: UIViewController {
-    
+ 
+    @objc public var delegate:OTRRoomOccupantsViewControllerDelegate? = nil
```
chrisballinger: this must be a weak var or you'll have a retain cycle

ChatSecure/Classes/View Controllers/OTRMessagesViewController.h
```diff
@@ -24,7 +24,7 @@
 
 @end
 
-@interface OTRMessagesViewController : JSQMessagesViewController <OTRMessagesViewControllerProtocol, UIPopoverPresentationControllerDelegate>
+@interface OTRMessagesViewController : JSQMessagesViewController <OTRMessagesViewControllerProtocol, UIPopoverPresentationControllerDelegate, OTRRoomOccupantsViewControllerDelegate>
```
chrisballinger: Please fix the warning here, probably by moving this to an internal class extension inside .m file. In the future you can filter warnings to 'chatsecurecore' as shown in the image below to avoid the noise from other dependencies.

screen shot 2017-09-14 at 2 32 30 pm

ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomManager.m
```diff
@@ -148,6 +154,7 @@ - (void)inviteBuddies:(NSArray<NSString *> *)buddyUniqueIds toRoom:(XMPPRoom *)r
     }];
     [buddies enumerateObjectsUsingBlock:^(OTRXMPPBuddy * _Nonnull buddy, NSUInteger idx, BOOL * _Nonnull stop) {
         [self inviteUser:buddy.bareJID toRoom:room.roomJID withMessage:nil];
+        [self setUser:buddy.bareJID affiliation:@"admin" inRoom:room.roomJID];
```
chrisballinger: I agree, I'm not sure if this is a good default for upstream because it may lead to unexpected behavior for more advanced XMPP users. We already have complaints that we are setting room parameters for every room we join, which I'm not sure if it's been fixed. However perhaps it could be a default if it was at least configurable.

I think there is still a use case for Zom for having a toggle to not make everyone an admin, for example a very large group chat with potentially less trusted invited members. I wonder if we could have an button on the invite view to push a "Group Options" view that would allow for additional configuration flexibility in the future. For now it could have the single option:

* **Trust Invited Contacts** subtitle: Invited members will be made room administrators. When creating a large room with less untrusted contacts, you may want to disable this. default: ON

The room options could be a serializable object (MTLModel subclass I guess) with a single boolean option `makeInviteesAdmin`. Then we could store the room options on the OTRXMPPRoom object as  `roomOptions`, and use that for this logic.

Then the below logic could look like:

```objc
OTRXMPPRoom *xmppRoom = ...
if (xmppRoom.roomOptions.makeInviteesAdmin) {
    [self setUser:buddy.bareJID affiliation:@"admin" inRoom:room.roomJID];
}
```

cc @cstiens

N-Pex: I agree, making all group members admin somehow feels... wrong. Let's discuss today.

chrisballinger:
This is now all merged into the master branch, with the one exception of reverting the "admin" commit until a better solution is thoroughly investigated.

PR #855, opened 2017-10-05
Small member list related fixes for groups

N-Pex:


ChatSecure/Classes/View Controllers/OTRComposeGroupViewController.swift
```diff
@@ -29,7 +29,7 @@ open class OTRComposeGroupViewController: UIViewController, UICollectionViewDele
     
     var selectedItems:[OTRXMPPBuddy] = []
     var prototypeCell:OTRComposeGroupBuddyCell?
-    var excludedItems:[String]? = nil
+    var existingItems:[String]? = nil
```
chrisballinger: In the future, it would probably be more appropriate to use a non-optional mutable Swift Set because it's being used for lookups. O(1) vs O(n)

```swift
var existingItems = Set<String>()
```

PR #861, opened 2017-10-20
Invalidate stored size info and "one message cache"

N-Pex:
This message cache thingy "currentMessageIndex" could have caused a looooot of strange issues, since it was never reset in "onReceivedChanges". This meant we could have hit a different object from the one we thought (the currentMessage) in a lot of places!

### DIFF
--- a/ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
+++ b/ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
@@ -2001,8 +2001,17 @@ heightForCellBottomLabelAtIndexPath:(NSIndexPath *)indexPath
             [alert addAction:obj];
         }];
         [alert addAction:[self cancleAction]];
+        
+        // Get the anchor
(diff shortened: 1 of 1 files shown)


chrisballinger:
Looks good, thank you!

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -2040,41 +2089,8 @@ - (void)didReceiveChanges:(OTRYapViewHandler *)handler sectionChanges:(NSArray<Y
         } else {
             // We can't use finishSendingMessage here because it might
             // accidentally clear out unsent message text
-            [self.collectionView.collectionViewLayout invalidateLayoutWithContext:[JSQMessagesCollectionViewFlowLayoutInvalidationContext context]];
```
chrisballinger: @N-Pex Is there any way that this line removal is causing the crashes?

PR #867, opened 2017-10-26
Trying to fix crashes in Message View Controller.

N-Pex:


chrisballinger:
I think you have some good ideas here, I'm going to merge this and play around with it more over the weekend.

PR #870, opened 2017-10-29
More reliable disconnects

stigger:
Disconnect from the background tasks does not always work correctly, because the background task terminates too early. As a result, on the next attempt to connect we're sometimes unable to resume the session.

### DIFF
--- a/ChatSecure/Classes/Controllers/OTRProtocolManager.m
+++ b/ChatSecure/Classes/Controllers/OTRProtocolManager.m
@@ -206,17 +206,32 @@
     }
 }
 
-- (void)disconnectAllAccountsSocketOnly:(BOOL)socketOnly {
+- (void)disconnectAllAccountsSocketOnly:(BOOL)socketOnly andWait:(BOOL)wait {
     @synchronized (self) {
-        [self.protocolManagers enumerateKeysAndObjectsUsingBlock:^(id key, id <OTRProtocol> protocol, BOOL *stop) {
-            [protocol disconnectSocketOnly:socketOnly];
-        }];
+        for (id<OTRProtocol> manager in self.protocolManagers.allValues) {
+            [manager disconnectSocketOnly:socketOnly];
+        }
+        if (wait) {
+            while (true) {
+                BOOL stillDisconnecting = NO;
+                for (id <OTRProtocol> manager in self.protocolManagers.allValues) {
+                    if (manager.connectionStatus != OTRProtocolConnectionStatusDisconnected) {
+                        stillDisconnecting = YES;
+                        break;
+                    }
+                }
+                if (!stillDisconnecting) {
+                    return;
+                }
+                [NSThread sleepForTimeInterval:.02];
+            }
+        }
     }
 }
 
 - (void)disconnectAllAccounts
 {
-    [self disconnectAllAccountsSocketOnly:NO];
+    [self disconnectAllAccountsSocketOnly:NO andWait:NO];
 }
 
 - (void)protocolDidChange:(NSDictionary *)change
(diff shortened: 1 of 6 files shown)


chrisballinger:
This is definitely a good idea and will probably fix some issues I've been having, but I'd like to try an alternative approach than sleeping the thread in an infinite loop. Although its a little more complicated, you could do KVO on each `connectionStatus` with a `dispatch_group` and wait timeout equal to slightly less than the background time remaining, and a completionBlock parameter that returns YES if it could disconnect all the accounts in time, or NO if it ran out of time.

```objc
(void)disconnectAllAccountsSocketOnly:(BOOL)socketOnly timeout:(NSTimeInterval)timeout completionBlock:(void (^)(BOOL success)) completionBlock;
```

stigger:
Thanks for the suggestion, implemented. However, I'm not passing the boolean parameter to the completion block, because it doesn't really matter whether the timeout has elapsed or not: even if it has, there is nothing we can do about it.

ChatSecure/Classes/Controllers/OTRProtocolManager.m
```diff
@@ -206,17 +207,37 @@ - (void)goAwayForAllAccounts {
...
+        for (NSObject<OTRProtocol> *manager in self.protocolManagers.allValues) {
+            if (manager.connectionStatus != OTRProtocolConnectionStatusDisconnected) {
+                dispatch_group_enter(group);
+                __block NSString *token = nil;
+                token = [manager addObserverForKeyPath:@"connectionStatus"
+                                               options:0
+                                                 block:^(NSString *keyPath, NSObject <OTRProtocol> *mgr, NSDictionary *change) {
+                                                     if (mgr.connectionStatus == OTRProtocolConnectionStatusDisconnected) {
+                                                         dispatch_group_leave(group);
+                                                         [mgr removeObserverForToken:token];
```
chrisballinger: This could cause an observer leak. You might want to try using Facebook's [`KVOController`](https://github.com/facebook/KVOController) which is already a dependency, because it makes KVO a little easier.

ChatSecure/Classes/Controllers/OTRProtocolManager.m
```diff
@@ -206,17 +207,37 @@ - (void)goAwayForAllAccounts {
...
     @synchronized (self) {
-        [self.protocolManagers enumerateKeysAndObjectsUsingBlock:^(id key, id <OTRProtocol> protocol, BOOL *stop) {
-            [protocol disconnectSocketOnly:socketOnly];
-        }];
+        dispatch_group_t group = dispatch_group_create();
+        for (NSObject<OTRProtocol> *manager in self.protocolManagers.allValues) {
+            if (manager.connectionStatus != OTRProtocolConnectionStatusDisconnected) {
+                dispatch_group_enter(group);
+                __block NSString *token = nil;
+                token = [manager addObserverForKeyPath:@"connectionStatus"
```
chrisballinger: Instead of using a string literal it's slightly better to use `NSStringFromSelector(@selector(connectionStatus))`

ChatSecure/Classes/Controllers/OTRProtocolManager.m
stigger: Sorry, don't see the leak. `mgr` here is a local parameter, not a captured variable.

chrisballinger: What I'm trying to say is that if the `connectionStatus` doesn't equal `OTRProtocolConnectionStatusDisconnected`, the observer won't be removed. If the dispatch group wait timer expires without the observers being removed.. things might get weird. You could keep track of the tokens and then remove any non-removed observers after the dispatch_group_wait.

stigger: In that case the disconnected status would be set next time the app switches to foreground and the observers would still be removed, but just to be safe and avoid confusion I changed it so all of them are always cleaned up in a single place.

chrisballinger:
Ok looks good, thanks for making those changes!

PR #872, opened 2017-10-31
WIP: Group chat fixes

N-Pex:
Try to use real buddy avatars and display names in group chats and some other small changes/additions. Should probably be tested more, but posted here as a WIP.

### DIFF
--- a/ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
+++ b/ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import YapDatabase
+import Mantle
 
 @objc public enum RoomOccupantRole:Int {
     case none = 0
@@ -64,11 +65,25 @@ open class OTRXMPPRoomOccupant: OTRYapDatabaseObject, YapDatabaseRelationshipNod
     @objc open var affiliation:RoomOccupantAffiliation = .none
 
     /**When given by the server we get the room participants reall JID*/
-    @objc open var realJID:String?
+    @objc open var realJID:String? {
+        didSet {
+            if let realJid = self.realJID, let roomUniqueId = self.roomUniqueId {
+                OTRDatabaseManager.shared.readOnlyDatabaseConnection?.asyncRead({ (transaction) in
+                    if let room = OTRXMPPRoom.fetchObject(withUniqueID: roomUniqueId, transaction: transaction), let accountUniqueId = room.accountUniqueId {
+                        self.realBuddy = OTRBuddy.fetch(withUsername: realJid, withAccountUniqueId: accountUniqueId, transaction: transaction)
+                    }
+                })
+            }
+        }
+    }
+    @objc open var realBuddy:OTRBuddy?
     
     @objc open var roomUniqueId:String?
     
     @objc open func avatarImage() -> UIImage {
+        if let buddy = self.realBuddy {
+            return OTRImages.avatarImage(withUniqueIdentifier: buddy.uniqueId, avatarData: nil, displayName: buddy.displayName, username: buddy.username)
+        }
         return OTRImages.avatarImage(withUniqueIdentifier: self.uniqueId, avatarData: nil, displayName: roomName ?? realJID ?? jid, username: self.realJID)
     }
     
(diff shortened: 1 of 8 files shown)


chrisballinger:
Thanks I'll review this and get back to you with suggestions

ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
```diff
@@ -64,11 +65,25 @@ open class OTRXMPPRoomOccupant: OTRYapDatabaseObject, YapDatabaseRelationshipNod
...
+            if let realJid = self.realJID, let roomUniqueId = self.roomUniqueId {
+                OTRDatabaseManager.shared.readOnlyDatabaseConnection?.asyncRead({ (transaction) in
+                    if let room = OTRXMPPRoom.fetchObject(withUniqueID: roomUniqueId, transaction: transaction), let accountUniqueId = room.accountUniqueId {
+                        self.realBuddy = OTRBuddy.fetch(withUsername: realJid, withAccountUniqueId: accountUniqueId, transaction: transaction)
+                    }
+                })
+            }
+        }
+    }
+    @objc open var realBuddy:OTRBuddy?
```
chrisballinger: Instead of doing database transactions in didSet and then assigning a whole database object, I think it would be better to:

1. Fetch the corresponding OTRBuddy in the YapStorage class
2. Just store the OTRBuddy.uniqueId on `OTRXMPPRoomOccupant.buddyUniqueId` as a `String?`.
3. Add an accessor function `func buddy(with: YapDatabaseReadTransaction) -> OTRXMPPBuddy?` that does a simple lookup.

ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
```diff
@@ -64,11 +65,25 @@ open class OTRXMPPRoomOccupant: OTRYapDatabaseObject, YapDatabaseRelationshipNod
...
     
     @objc open var roomUniqueId:String?
     
     @objc open func avatarImage() -> UIImage {
+        if let buddy = self.realBuddy {
+            return buddy.avatarImage
+        }
```
chrisballinger: You can this logic elsewhere (see below)

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -1632,13 +1647,15 @@ - (NSString *)senderDisplayName
...
-                roomOccupant = occupant;
-                *stop = YES;
-            }];
+        [self.readOnlyDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            roomOccupant = [self occupantWithTransaction:transaction forFullJid:roomMessage.senderJID inRoom:roomMessage.roomUniqueId];
         }];
-        UIImage *avatarImage = [roomOccupant avatarImage];
+        UIImage *avatarImage = nil;
+        if (roomOccupant) {
+            avatarImage = [roomOccupant avatarImage];
```
chrisballinger: Above you can fetch the realBuddy using the func, and then just use the avatarImage of the buddy if it's available.

ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
chrisballinger: I still think having an implicit database transaction here is bad news. If you accidentally have a transaction inside another transaction you'll likely deadlock. Either have a function where you pass in a transaction, or handle the logic to determine which avatar to return at a higher level.

N-Pex: Yeah, I agree, I realized late yesterday that's probably what you meant, but didn't have time to fix it. Done now! Spent most of the day trying to get things to build, after Swift 4 update. I could not find ZXingObjC 3.2.2 (maybe need to update something?) but figured out I could use your xcode9 branch for now.

chrisballinger: You'll need to update the Carthage dependencies:

```
$ rm -rf ~/Library/Caches/org.carthage.CarthageKit
$ carthage update --platform ios --cache-builds
```

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -358,6 +360,21 @@ - (nullable OTRAccount *)accountWithTransaction:(nonnull YapDatabaseReadTransact
     return account;
 }
 
+- (nullable OTRXMPPRoomOccupant *)occupantWithTransaction:(nonnull YapDatabaseReadTransaction *)transaction forFullJid:(nonnull NSString*)fullJid inRoom:(nonnull NSString *)roomUniqueId {
```
chrisballinger: This would be a lot faster with a secondary index. Right now it's basically O(n) for the number of occupants. Considering this method will be called repeatedly while the UI is scrolling around (and nothing is cached), I would not be surprised to see noticeable performance problems / stuttering for larger rooms. Also I think this helper function would be better moved to the `OTRXMPPRoomOccupant` or `OTRXMPPRoom` classes.

Admittedly it's a bit of a mess to set up the secondary indexes so I can handle this myself after the merge. Just something to note for future consideration, try to avoid blocking O(n) lookups in frequently called UI functions.

PR #873, opened 2017-10-31
865 fix

afriedmanGlacier:
I'm not sure exactly what you mean by a "YapDatabase relationship deletion rule" but I tried to make fixes for deleting images and for removing user from group when thread is deleted. Haven't had any issues so far in testing these over a few days.

I apologize, I have no idea why the first commit is part of this pull request. I was just trying to get the latest from your master. Not sure how to remove that commit from the pull request.

### DIFF
--- a/ChatSecure/Classes/Controllers/OTRMediaFileManager.h
+++ b/ChatSecure/Classes/Controllers/OTRMediaFileManager.h
@@ -30,6 +30,12 @@ extern NSString *const kOTRRootMediaDirectory;
      completion:(void (^)(NSInteger bytesWritten, NSError * _Nullable error))completion
 completionQueue:(nullable dispatch_queue_t)completionQueue;
 
+//#865
+- (void)deleteDataForItem:(OTRMediaItem *)mediaItem
+            buddyUniqueId:(NSString *)buddyUniqueId
+               completion:(void (^)(BOOL success, NSError * _Nullable error))completion
+          completionQueue:(nullable dispatch_queue_t)completionQueue;
+
 - (nullable NSData*)dataForItem:(OTRMediaItem *)mediaItem
                   buddyUniqueId:(NSString *)buddyUniqueId
                           error:(NSError* __autoreleasing *)error;
--- a/ChatSecure/Classes/Controllers/OTRMediaFileManager.m
+++ b/ChatSecure/Classes/Controllers/OTRMediaFileManager.m
@@ -128,6 +128,50 @@ completionQueue:(nullable dispatch_queue_t)completionQueue {
     }];
 }
 
+//#865
+- (void)deleteDataForItem:(OTRMediaItem *)mediaItem
+            buddyUniqueId:(NSString *)buddyUniqueId
+               completion:(void (^)(BOOL success, NSError * _Nullable error))completion
+          completionQueue:(nullable dispatch_queue_t)completionQueue {
+    if (!completionQueue) {
+        completionQueue = dispatch_get_main_queue();
+    }
+    [self performAsyncWrite:^{
+        NSString *path = [[self class] pathForMediaItem:mediaItem buddyUniqueId:buddyUniqueId];
+        if (![path length]) {
+            NSError *error = [NSError errorWithDomain:kOTRErrorDomain code:150 userInfo:@{NSLocalizedDescriptionKey:@"Unable to create file path"}];
+            if (completion) {
+                dispatch_async(completionQueue, ^{
+                    completion(NO, error);
+                });
+            }
+            return;
+        }
+        
+        BOOL fileExists = [self.ioCipher fileExistsAtPath:path isDirectory:NULL];
+        
+        if (fileExists) {
+            NSError *error = nil;
+            [self.ioCipher removeItemAtPath:path error:&error];
+            if (error) {
+                NSError *error = [NSError errorWithDomain:kOTRErrorDomain code:151 userInfo:@{NSLocalizedDescriptionKey:@"Unable to remove existing file"}];
+                if (completion) {
+                    dispatch_async(completionQueue, ^{
+                        completion(NO, error);
+                    });
+                }
+                return;
+            }
+        }
+        
+        if (completion) {
+            dispatch_async(completionQueue, ^{
+                completion(YES, nil);
+            });
+        }
+    }];
+}
+
 /* Internal. If "length" is set, only return the length of the data, otherwise the data ifself */
 - (nullable NSData*)dataForItem:(OTRMediaItem *)mediaItem
                   buddyUniqueId:(NSString *)buddyUniqueId
--- a/ChatSecure/Classes/Model/Yap Storage/OTRMediaItem.m
+++ b/ChatSecure/Classes/Model/Yap Storage/OTRMediaItem.m
@@ -329,7 +329,29 @@ static NSString* GetExtensionForMimeType(NSString* mimeType) {
 
 - (id)yapDatabaseRelationshipEdgeDeleted:(YapDatabaseRelationshipEdge *)edge withReason:(YDB_NotifyReason)reason
 {
-    //TODO:Delete File because the parent OTRMessage was deleted
+    //#865 Delete File because the parent OTRMessage was deleted
+    __block id<OTRThreadOwner> thread = nil;
+    __block id<OTRMessageProtocol> message = nil;
+    [OTRDatabaseManager.shared.readOnlyDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+        message = [self parentMessageWithTransaction:transaction];
+        thread = [message threadOwnerWithTransaction:transaction];
+    }];
+    if (!message || !thread) {
+        DDLogError(@"Missing parent message or thread for media message!");
+        return nil;
+    }
+    
+    NSString *buddyUniqueId = [thread threadIdentifier];
+    if (!buddyUniqueId) {
+        return nil;
+    }
+    
+    [[OTRMediaFileManager sharedInstance] deleteDataForItem:self buddyUniqueId:buddyUniqueId completion:^(BOOL success, NSError *error) {
+        if (error) {
+            DDLogError(@"ERROR in deleting data for media item");
+        }
+    } completionQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)];
+    
     return nil;
 }
 
--- a/Podfile
+++ b/Podfile
@@ -59,8 +59,8 @@ abstract_target 'ChatSecureCorePods' do
   pod 'gtm-oauth2', :podspec => 'Podspecs/gtm-oauth2.podspec'
 
   # Storage
-  pod 'YapDatabase/SQLCipher', '~> 3.0.2'
-  # pod 'YapDatabase/SQLCipher', :git => 'https://github.com/yapstudios/YapDatabase.git', :commit => 'd632658'
+  # pod 'YapDatabase/SQLCipher', '~> 3.0.2'
+  pod 'YapDatabase/SQLCipher', :git => 'https://github.com/yapstudios/YapDatabase.git', :commit => 'ce9c8db' # 865 fix
   pod 'libsqlfs/SQLCipher', :git => 'https://github.com/ChatSecure/libsqlfs.git', :branch => 'podspec-fix'
   pod 'IOCipher/GCDWebServer', :path => 'Submodules/IOCipher/IOCipher.podspec'
   pod 'YapTaskQueue/SQLCipher', :git => 'https://github.com/ChatSecure/YapTaskQueue.git', :branch => 'swift4'


chrisballinger:
Thanks! Would you mind signing the [CLA](https://github.com/ChatSecure/ChatSecure-iOS/blob/master/CONTRIBUTING.md) before I merge?

ChatSecure/Classes/Controllers/OTRMediaFileManager.h
```diff
@@ -30,6 +30,12 @@ extern NSString *const kOTRRootMediaDirectory;
      completion:(void (^)(NSInteger bytesWritten, NSError * _Nullable error))completion
 completionQueue:(nullable dispatch_queue_t)completionQueue;
 
+//#865
+- (void)deleteDataForItem:(OTRMediaItem *)mediaItem
+            buddyUniqueId:(NSString *)buddyUniqueId
+               completion:(void (^)(NSError * _Nullable error))completion
```
chrisballinger: Please use `BOOL success, NSError * _Nullable error`

ChatSecure/Classes/Controllers/OTRMediaFileManager.m
```diff
@@ -128,6 +128,44 @@ - (void)setData:(NSData *)data
...
+          completionQueue:(nullable dispatch_queue_t)completionQueue {
+    if (!completionQueue) {
+        completionQueue = dispatch_get_main_queue();
+    }
+    [self performAsyncWrite:^{
+        NSString *path = [[self class] pathForMediaItem:mediaItem buddyUniqueId:buddyUniqueId];
+        if (![path length]) {
+            NSError *error = [NSError errorWithDomain:kOTRErrorDomain code:150 userInfo:@{NSLocalizedDescriptionKey:@"Unable to create file path"}];
+            dispatch_async(completionQueue, ^{
+                completion(error);
```
chrisballinger: Check if completion is nil

ChatSecure/Classes/Controllers/OTRMediaFileManager.m
```diff
@@ -128,6 +128,44 @@ - (void)setData:(NSData *)data
...
+        
+        BOOL fileExists = [self.ioCipher fileExistsAtPath:path isDirectory:NULL];
+        
+        if (fileExists) {
+            NSError *error = nil;
+            [self.ioCipher removeItemAtPath:path error:&error];
+            if (error) {
+                NSError *error = [NSError errorWithDomain:kOTRErrorDomain code:151 userInfo:@{NSLocalizedDescriptionKey:@"Unable to remove existing file"}];
+                dispatch_async(completionQueue, ^{
+                    completion(error);
```
chrisballinger: Check if completion is nil

ChatSecure/Classes/Controllers/OTRMediaFileManager.m
```diff
@@ -128,6 +128,44 @@ - (void)setData:(NSData *)data
...
+                });
+                return;
+            }
+        }
+        
+        dispatch_async(completionQueue, ^{
+            completion(nil);
```
chrisballinger: Check if completion is nil

ChatSecure/Classes/Controllers/YapDatabaseTransaction+ChatSecure.swift
```diff
@@ -107,6 +107,29 @@ public extension YapDatabaseReadTransaction {
...
+    
+    //#865
+    @objc public func allMediaMessagesForThread(_ thread:OTRThreadOwner) -> [OTRMessageProtocol] {
+        guard let indexTransaction = self.ext(OTRMessagesSecondaryIndex) as? YapDatabaseSecondaryIndexTransaction else {
+            return []
+        }
+        let queryString = "Where \(OTRYapDatabaseMessageThreadIdSecondaryIndexColumnName) == ?"
+        let query = YapDatabaseQuery(string: queryString, parameters: [thread.threadIdentifier])
+        var result = [OTRMessageProtocol]()
+        let success = indexTransaction.enumerateKeysAndObjects(matching: query) { (collection, key, object, stop) in
```
chrisballinger: Instead of iterating over every message, it would be a faster query to add `messageMediaItemKey` to the secondary index columns (see setupSecondaryIndexes in OTRDatabaseManager).

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
```diff
@@ -251,6 +251,21 @@ + (void)deleteAllMessagesWithTransaction:(YapDatabaseReadWriteTransaction*)trans
...
+    mediaMessages = [transaction allMediaMessagesForThread:threadOwner];
+    [mediaMessages enumerateObjectsUsingBlock:^(id<OTRMessageProtocol>  _Nonnull medmsg, NSUInteger idx, BOOL * _Nonnull stop) {
+        item = [OTRMediaItem mediaItemForMessage:medmsg transaction:transaction];
+        [[OTRMediaFileManager sharedInstance] deleteDataForItem:item buddyUniqueId:uniqueBuddyId completion:^(NSError *error) {
+            if (error) {
+                DDLogError(@"ERROR in deleting data for media item");
+            }
+        } completionQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)];
+    }];
+
```
chrisballinger: This can be improved by using YapDatabase deletion notify in both OTRBaseMessage and OTRXMPPRoomMessage, or perhaps OTRMediaItem: https://github.com/yapstudios/YapDatabase/wiki/Relationships#notify

```objc
- (nullable id)yapDatabaseRelationshipEdgeDeleted:(YapDatabaseRelationshipEdge *)edge
                                       withReason:(YDB_NotifyReason)reason {
// delete the data here
}
```

Normally Yap will automatically delete child objects (e.g. messages are deleted when a buddy is deleted) if the relationships are setup properly. In theory this can be chained down to the OTRMediaItem, and then `[[OTRMediaFileManager sharedInstance] deleteDataForItem` can be called inside `yapDatabaseRelationshipEdgeDeleted:withReason:` for each item individually.

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
```diff
@@ -270,6 +285,36 @@ + (void)deleteAllMessagesForAccountId:(NSString *)uniqueAccountId transaction:(Y
     [[transaction ext:extensionName] enumerateEdgesWithName:edgeName destinationKey:uniqueAccountId collection:[OTRAccount collection] usingBlock:^(YapDatabaseRelationshipEdge *edge, BOOL *stop) {
         [self deleteAllMessagesForBuddyId:edge.sourceKey transaction:transaction];
     }];
+    
+    //#865
+    [self removeAccountFromGroups:uniqueAccountId transaction:transaction];
```
chrisballinger: This might add unexpected behavior. The current setting to delete messages on disconnection doesn't delete your roster, and rooms are sort of your roster. I think this should be extracted and discussed further in a separate issue.

ChatSecure/Classes/Controllers/YapDatabaseTransaction+ChatSecure.swift
afriedmanGlacier: I'm struggling with this one, can't quite figure out how this is supposed to work. Our input parameter to this function is OTRThreadOwner, and we are checking all the messages in a thread to see if they are media messages. Would the new secondary column then be in setupMessageSecondaryIndexes instead of setupSecondaryIndexes?

Also, we are just checking the existence of mediaMessages here, so should we just query for the existence of data in this column? Something like:

let queryString = "Where \(OTRYapDatabaseMessageThreadIdSecondaryIndexColumnName) == ? AND \(OTRYapDatabaseMessageMediaItemKeySecondaryIndexColumnName) != NULL"

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
afriedmanGlacier: In this case, then the allMediaMessagesForThread question above is irrelevant and we don't need that function, right?

afriedmanGlacier: In OTRBaseMessage.yapDatabaseRelationshipEdges, it appears to create the relationship for OTRMediaItems with nodeDeleteRules:YDB_DeleteDestinationIfSourceDeleted | YDB_NotifyIfSourceDeleted. 

But yapDatabaseRelationshipEdgeDeleted in OTRMediaItem doesn't get called when I delete a thread with media items. I'm not sure why. yapDatabaseRelationshipEdgeDeleted in OTRMediaItem doesn't call deleteDataForItem yet, but I just wanted to see if the method got called. It seems like it would be better to use yapDatabaseRelationshipEdgeDeleted in OTRMediaItem, but since I'm not sure why it isn't getting called, maybe I'll try in OTRBaseMessage and OTRXMPPRoomMessage

afriedmanGlacier: The yapDatabaseRelationshipEdgeDeleted selector is never getting called from the flush method in YapDatabaseRelationshipTransaction. Was tracking down why, but had to do other things and ran out of time

chrisballinger: Thanks for your work here, I think you are pretty close and just need to play around with the nodeDeleteRules (perhaps you want destination not source?). You can repurpose some of the existing tests to speed things up if you'd like. If you see FileTransferIntegrationTests.swift you'll see an example of how to set up a test database. Just make sure when making your test data that you don't have an orphan object or it'll get deleted out from under you. (e.g. make Account, Buddy, and Message with proper relationships).. see FileTransferIntegrationTests.setUp()

afriedmanGlacier: OTRBaseMessage currently has YDB_DeleteDestinationIfSourceDeleted, which deletes the media item (in YapDatabaseRelationshipTransaction) without sending notification. 

If we change it to YDB_NotifyIfSourceDeleted, it will send the notification and call yapDatabaseRelationshipEdgeDeleted in OTRMediaItem (where we could remove the file with deleteDataForItem), but it doesn't actually delete the destination edge item in this case. Returning nil does nothing, otherwise you can return a modified destination object which will update the database

So, I'm thinking there are a few possibilities, but maybe I'm missing something. It seems to me like YapDatabaseRelationshipTransaction should do one of the following:

1) Another nodeDeleteRule for YDB_DeleteDestinationAndNotifyIfSourceDeleted (same for source).
2) Have the option to delete if user returns nil from yapDatabaseRelationshipEdgeDeleted. However, this would mean any user would need to return the same, unmodified object if they DON'T want to delete in this case.

The flush method is already huge with all the conditions, but having the option to Notify and Delete would be helpful in this case.

All that being said, the current fix that I proposed in the original pull request is inefficient (as you said in your first comments), but it works. Sooooo...should we just hold off on the whole thing for now and just put a change request in for the YapDatabase pod?

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
afriedmanGlacier: With this one, are you saying I should create a separate ticket for handling Groups when auto-deleting on disconnect or manually deleting a group thread? Or just shelf it for now and think about what the expected behavior should be?

chrisballinger: Yeah I think that should be separated into a different issue for further discussion so it doesn't hold up this PR.

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
chrisballinger: Oh interesting, so the "proper fix" is blocked on a YapDatabase bug? If that's the case, definitely open an issue on YapDatabase describing the problem we're seeing here, and I'll reach out to Robbie to see if he has any thoughts before we issue a pull request. 

I'd prefer to do it right the first time, because this will allow you to properly delete individual messages, which is something people have asked about for a long time.

afriedmanGlacier: I just realized one more thing I need to test based on their example. I might be wrong and we might just need to change the rule. If that test fails, then I'll post on YapDatabase.

afriedmanGlacier: My other test failed. Here's the ticket I submitted: https://github.com/yapstudios/YapDatabase/issues/426

chrisballinger: Thank you!

chrisballinger:
@afriedmanGlacier We should submit a pull request to YapDatabase to get this moving along, and I'll do what I can to get it merged upstream.

afriedmanGlacier:
Ok, sounds good. I’ll try to get to that later this week or early next week.

ChatSecure/Classes/Model/Yap Storage/OTRBaseMessage.m
afriedmanGlacier: Looking at this again, I am going around in circles (and making a mess) as to the best way to handle this. The changes I started to make in YapDatabaseRelationshipTransaction would work for this ChatSecure situation but may not fit with their intentions for the code. Inside the YDB_DeleteDestinationIfSourceDeleted condition (and YDB_DeleteSourceIfDestinationDeleted), I was planning to check for the YDB_NotifyIfSourceDeleted condition and notify before deleting in order to allow preprocessing (removing the media item file at its path on disk). However, the NotifyIf conditions allow the listener to modify the object and send it back. That wouldn't do any good here because after the notification, we are deleting the object. If the client code modified it, it would then just get deleted. I'm probably not explaining that well.

I'm still not sure why delete and notify doesn't work the way it says at https://github.com/yapstudios/YapDatabase/wiki/Relationships#notify. 

One other option might be for us to leave YapDatabase alone (though I still think it isn't doing what its intended to do), and just put YDB_NotifyIfSourceDeleted for media items. This way, the notification works correctly and we can remove the file from disk, and then from that method call a thread to delete the mediaItem object. Thoughts?

afriedmanGlacier:
@chrisballinger the fix to YapDatabase by Robbie looks good. Guessing we should wait until its in their master build in order to finish the pull request based on their update?

chrisballinger:
im ok w/ pinning to the fixed commit in the podfile for now

afriedmanGlacier:
I'm sorry, I have no idea what just happened. I was trying to clean up the branch in order to resubmit it, but whatever I did automatically closed this. Sigh...this will be good in the long run as I need to get more familiar with git, but its a painful process at the moment.

afriedmanGlacier:
The changes are made and I think its ready.

PR #886, opened 2017-11-14
Group name and avatar fix

N-Pex:
Use new method to get occupant, remove "displayName" from room message object.


ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomYapStorage.m
```diff
@@ -171,43 +164,34 @@ - (BOOL)configureWithParent:(XMPPRoom *)aParent queue:(dispatch_queue_t)queue
...
-                    buddyRole = [item attributeStringValueForName:@"role"];
-                    buddyAffiliation = [item attributeStringValueForName:@"affiliation"];
-                    *stop = YES;
-                }
-            }];
-            *stop = YES;
-        }
-    }];
+    
+    DDXMLElement *item = nil;
```
chrisballinger: In the future please use `NSXMLElement` instead of `DDXMLElement`

PR #898, opened 2017-11-29
Don't auto download media for non-friends in group chat

N-Pex:
Also, show an info message in the chat view.


ChatSecure/Classes/View Controllers/OTRMessagesCollectionViewFlowLayout.swift
```diff
@@ -27,4 +43,41 @@ import JSQMessagesViewController
...
+        }
+        return size
+    }
+    
+    override open func layoutAttributesForElements(in rect: CGRect) -> [UICollectionViewLayoutAttributes]? {
+        guard var attributes = super.layoutAttributesForElements(in: rect) else { return nil }
+        
+        var offset:CGFloat = 0
+        for attribute in attributes {
+            if (attribute.representedElementCategory == UICollectionElementCategory.cell) {
```
chrisballinger: you can use a where clause in the for loop instead of inner if

PR #907, opened 2017-12-12
Supplementary view layout fixes

N-Pex:


chrisballinger:
Although I can confirm that this fixes the layout issues, there is another crash related to the supplementary view when refreshing a cell from a manual download:

```objc
                case YapDatabaseViewChangeUpdate :
                {
                    // Update could be e.g. when we are done auto-loading a link. We
                    // need to reset the stored size of this item, so the image/message
                    // will get the correct bubble height.
                    id <JSQMessageData> message = [self messageAtIndexPath:rowChange.indexPath];
                    [self.collectionView.collectionViewLayout.bubbleSizeCalculator resetBubbleSizeCacheForMessageData:message];
                    [self.messageSizeCache removeObjectForKey:@(message.messageHash)];
                    [self.collectionView reloadItemsAtIndexPaths:@[ rowChange.indexPath]];
                    break;
                }
```

I think this code is missing something. If you cherry-pick this commit https://github.com/ChatSecure/ChatSecure-iOS/commit/c82e41a133771ab6f7a61ac2c56ff5049c1ea2d4 which fixes the download button doing nothing, when the download is finished and tries to refresh the cell, it crashes.

Maybe it's because the cell type is changing from the placeholder to an image cell? Something needs to be invalidated?

It crashes in 

```swift
    open override func layoutAttributesForSupplementaryView(ofKind elementKind: String, at indexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
        let cacheIndex = cacheIndexFor(item: indexPath.item, section: indexPath.section)
        if let views = self.supplementaryViews[cacheIndex], let itemAttrs = self.layoutAttributesForItem(at: indexPath) {
            var viewOffset:CGFloat = 0
            for view in views {
                if view.kind == elementKind {
                    let suppViewAttrs = UICollectionViewLayoutAttributes(forSupplementaryViewOfKind: view.kind, with: indexPath)
                    suppViewAttrs.frame = CGRect(x: itemAttrs.frame.origin.x, y: itemAttrs.frame.origin.y+itemAttrs.frame.size.height+viewOffset, width: itemAttrs.frame.size.width, height: view.height)
                    return suppViewAttrs
                }
                viewOffset += view.height
            }
        }
        return super.layoutAttributesForSupplementaryView(ofKind: elementKind, at: indexPath)
    }
```

because it can't get a "section" from the indexPath, which seems absurd. If you call "isEmpty" on the IndexPath it returns true. I tried adding a guard clause and checking for indexPAth.isEmpty but returning nil and returning super both still led to a crash deep within UICollectionView.

PR #908, opened 2017-12-14
Always return supplementary view attributes, even if empty

N-Pex:


chrisballinger:
Unfortunately this doesn't fix the crash, and it now crashes as soon as I reach the supplementary view when scrolling up through the chat history. Of course Xcode gives a very unhelpful stack trace.

screen shot 2017-12-14 at 12 52 01 pm

chrisballinger:
Okay after attempting to reproduce from scratch it looks like it DID fix the crash. No idea where this other crash was coming from and I can't seem to reproduce it anymore.

PR #927, opened 2018-01-10
Subscription status

N-Pex:
Show "blocked" for contacts that have removed us. Also, don't allow sending to these contacts. I don't know how you want to handle this scenario upstream, let's discuss =)


chrisballinger:
I'm not sure if I can merge this as-is because it breaks the spec. Some things that need to be addressed:

1. In theory you should still be able to send messages to a contact even if you aren't mutually subscribed. The spec allows this behavior, and certain clients may depend on it, for instance [jmp.chat](https://jmp.chat).
2. The "right way" to do blocking is specified in [XEP-0191: Blocking Command](https://xmpp.org/extensions/xep-0191.html), but I believe in that scenario the blocked user cannot tell if they are blocked. This also requires adding some basic support in XMPPFramework.
3. There are cases where client or server bugs cause the subscription to be non-mutual, so there needs to be an easy way to trigger another subscription request.
4. I'm not sure "Blocked" is the right language to use here for a few reasons. If a user is harassing another user and know they are blocked, they can simply create new accounts and spam someone with presence requests.
5. I think adding a "subscription" string to the buddy model isn't exactly what we want. At the very least we should use an enum. 


While we're at it we might want another complimentary value, to allow having OTRXMPPBuddy objects in our local database that aren't synchronized with our server-side roster. This would allow receiving direct messages from users not on your roster, which we currently ignore. It would also allow us to keep around the temporary buddies that are created from non-anonymous RoomOccupants. To avoid spam we should still ignore messages from users not on your roster by default, or only keep them in the "archived" section.

N-Pex:
This is still work in progress, we need to flesh out the trustLevel a bit more and figure out where to set it to what...

ChatSecure/Classes/Model/Yap Storage/OTRXMPPBuddy.h
```diff
@@ -11,17 +11,31 @@
...
+@property (nonatomic) enum OTRXMPPBuddyTrustLevel trustLevel;
+@property (nonatomic) SubscriptionAttribute subscription;
+@property (nonatomic) SubscriptionPendingAttribute pending;
+
+- (BOOL) subscribedTo;
+- (BOOL) subscribedFrom;
+- (BOOL) pendingApproval;
+- (void) setPendingApproval:(BOOL)pending;
+- (BOOL) askingForApproval;
+- (void) setAskingForApproval:(BOOL)asking;
```
chrisballinger: These should be @properties, or better yet, defined in a Swift extension.

```swift
public extension OTRXMPPBuddy {
@objc public var subscribedTo: Bool {
get {
return blah
}
set {
do something with `newValue`
} 
}
```

ChatSecure/Classes/Model/Yap Storage/OTRXMPPBuddy.m
```diff
@@ -75,8 +103,36 @@ - (nullable XMPPJID*) bareJID {
...
+}
+
+- (BOOL) askingForApproval {
+    return [SubscriptionPendingAttributeBridge isPendingIn:self.pending];
+}
+
+- (void) setAskingForApproval:(BOOL)asking {
+    self.pending = [SubscriptionPendingAttributeBridge setPendingIn:self.pending pending:asking];
+}
+
```
chrisballinger: Move this to Swift extension described above

ChatSecure/Classes/Controllers/OTRDatabaseManager.m
```diff
@@ -130,7 +130,13 @@ - (BOOL)setupYapDatabaseWithName:(NSString *)name
     
     self.database = [[YapDatabase alloc] initWithPath:databasePath
                                            serializer:nil
-                                         deserializer:nil
+                                         deserializer:^ id (NSString *collection, NSString *key, NSData *data){
+                                             if ([collection isEqualToString:@"OTRXMPPPresenceSubscriptionRequest"]) {
```
chrisballinger: I'm not sure if this is what we should be doing here. I think we should keep the `OTRXMPPPresenceSubscriptionRequest` object around (stubbed out) so the deserialization operation doesn't fail, but then do a better job of making sure that we aren't doing things that would load objects from that collection in the first place.

We did something similar for the `OTRMessage` vs `OTRBaseMessage` migration.

chrisballinger:
Here are my notes / thoughts on trust level. Our old model worked under these assumptions / limitations:

1. The OTRBuddy collection tried to match your roster exactly.
2. Having someone in your OTRBuddy collection implied you trusted them to send you messages, see your presence, etc.
3. Incoming subscription requests would not create OTRBuddy objects and used a separate object because of (2).
4. We frequently would have to create temporary OTRBuddy objects to facilitate UI views that were populated by OTRBuddy model data. Care had to be made to not accidentally save these to the database.

By adding a `trustLevel` to the model, we want to allow the following behavior:

1. Allow storing objects in the OTRBuddy collection that are not necessarily in your roster. 
2. The main use cases are: room occupants, blocked buddies, non-roster buddies, (other?)
3. The other benefit is that we could allow users to receive messages from people not on their roster, which is part of the XMPP spec. Some use cases depend on this behavior, for example https://jmp.chat
4. I think we still don't want users to be able to manually add buddies that aren't on their roster.
5. Receiving messages from untrusted buddies should not show notifications and bypass the inbox and go straight into the "Archive" tab.

For now I suggest the following enum:

```obj-c
typedef NS_ENUM(NSInteger, BuddyTrustLevel) {
    /** default trust level */
    BuddyTrustLevelUntrusted = 0,
    /** this buddy is on the user's roster */
    BuddyTrustLevelRoster
};
```

This enum still doesn't help distinguish cases where you have a blocked buddy or they are semi-trusted by being a group chat participant. Perhaps an NS_OPTIONS bit flag would be better, but I don't think it maps perfectly either. I want to avoid adding the `BuddyTrustLevelTrusted` case for now in favor of just distinguishing between models in your roster vs not in your roster.

N-Pex:
Ok, I think I am "done" with this PR now =) Tested scanning and migrating today, both seemed to work after my latest commit.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
```diff
@@ -1383,6 +1383,56 @@ - (void) addBuddy:(OTRXMPPBuddy *)newBuddy
...
+        // We have an incoming subscription request, just answer that!
+        // TODO - use queue for this as well!
+        [buddy setAskingForApproval:NO];
+        [buddy saveWithTransaction:transaction];
+        
+        [self.xmppRoster acceptPresenceSubscriptionRequestFrom:jid andAddToRoster:YES];
+        
+        // Cleanup of old code. We no longer use the OTRXMPPPresenceSubscriptionRequest class,
+        // so if we have an old entry for this JID, just delete it
+        OTRXMPPPresenceSubscriptionRequest *presenceRequest = [OTRXMPPPresenceSubscriptionRequest fetchPresenceSubscriptionRequestWithJID:[jid bare] accontUniqueId:self.account.uniqueId transaction:transaction];
```
chrisballinger: You could also just do a `[transaction removeAllObjectsFromCollection: OTRXMPPPresenceSubscriptionRequest.collection]` during the database setup.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
```diff
@@ -1035,13 +1036,15 @@ - (void)parseSubscriptionFromRosterItem:(NSXMLElement *)item {
                 buddy.trustLevel = BuddyTrustLevelUntrusted;
                 buddy.displayName = [jid user];
                 buddy.username = [jid bare];
+            } else {
+                buddy = [buddy copy];
             }
             buddy.subscription = [SubscriptionAttributeBridge subscriptionWithString:subscription];
             [buddy setPendingApproval:[ask isEqualToString:@"subscribe"]];
```
chrisballinger: I thought pendingApproval could be readonly and calculated from the new enum cases we added?

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
```diff
@@ -1035,13 +1036,15 @@ - (void)parseSubscriptionFromRosterItem:(NSXMLElement *)item {
...
             if ([buddy subscribedFrom]) {
                 [buddy setAskingForApproval:NO];
             }
-            if ([buddy subscribedTo]) {
+            if ([buddy subscribedTo] || [buddy subscribedFrom]) {
                 buddy.trustLevel = BuddyTrustLevelRoster;
```
chrisballinger: Won't this add _any_ incoming roster items? This delegate callback `- (void)xmppRoster:(XMPPRoster *)sender didReceiveRosterItem:(NSXMLElement *)item;` from XMPPRosterDelegate doesn't distinguish between items that you've added to your roster and items that are incoming presence requests. The `XMPPRosterStorage` version of this callback `- (void)handleRosterItem:(NSXMLElement *)item xmppStream:(XMPPStream *)stream;` filters out non-roster items .

See `isRosterItem:` inside XMPPRoster.m for more details: 

```objc
/**
 * Some server's include in our roster the JID's of user's NOT in our roster.
 * This happens when another user adds us to their roster, and requests permission to receive our presence.
 * 
 * As discussed in RFC 3921, the state of the other user is "None + Pending In",
 * and the server "SHOULD NOT" include these JID's in the roster it sends us.
 * 
 * Nonetheless, some servers do anyway.
 * This method filters out such rogue entries in our roster.
 * 
 * Note that the server will automatically send us the proper presence subscription request,
 * and it will continue to do so everytime we sign in.
 * From the RFC:
 *     the user's server MUST keep a record of the subscription request and deliver the request when the
 *     user next creates an available resource, until the user either approves or denies the request.
 * 
 * So there is absolutely NO reason to process these entries, or include them in the roster's storage.
 * Furthermore, it isn't reliable to depend on these entires being there.
 * The RFC has clearly defined recommendations on the matter, and servers that currently send these rogue items
 * may very likely stop doing so in future versions.
**/
- (BOOL)isRosterItem:(NSXMLElement *)item
{
	NSString *subscription = [item attributeStringValueForName:@"subscription"];
	if ([subscription isEqualToString:@"none"])
	{
		NSString *ask = [item attributeStringValueForName:@"ask"];
		if ([ask isEqualToString:@"subscribe"])
		{
			return YES;
		}
		else
		{
			return NO;
		}
	}
	
	return YES;
}
```

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
N-Pex: No, since we split the 9 states into two separate fields, subscription { none, from, to, both } and pending { none, out, in, outin }.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
N-Pex: Yeah, we can discuss at which point we should set it to TrustLevelRoster, but I think we want the method to handle ALL incoming items, I am not a fan of the isRosterItem method, seems like a hack to me.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
N-Pex: Yeah, seems cleaner.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
chrisballinger: Ah I see but it's still stored internally using the new enums. I was confused and thought we still had this old bool floating around.

ChatSecure/Classes/Controllers/XMPP/OTRXMPPManager.m
chrisballinger: Hm, yeah.

Whatever we do I think this code should be unified with the RosterStorage code so we don't have two code paths doing different things. Right now rosterItems will flow from the XMPPRoster and end up being processed in two places (or sometimes one, depending on the result of isRosterItem)

```objc
/// inside OTRXMPPManager.m
- (void)xmppRoster:(XMPPRoster *)sender didReceiveRosterItem:(NSXMLElement *)item;
/// inside OTRYapDatabaseRosterStorage.m
- (void)handleRosterItem:(NSXMLElement *)item xmppStream:(XMPPStream *)stream;
```

Coudn't `- (void)parseSubscriptionFromRosterItem:(NSXMLElement *)item` be moved to `OTRYapDatabaseRosterStorage.m`? You could still make an `OTRYapDatabaseRosterStorage: XMPPRosterDelegate` extension and leave the rest of the vCard code inside XMPPManager, since the multicast delegate just lets you add as many delegates as you want.

```objc
// you can have both be delegates
	[self.xmppRoster addDelegate:self delegateQueue:self.workQueue];
	[self.xmppRoster addDelegate:self.xmppRosterStorage delegateQueue:self.workQueue];
```

chrisballinger:
@N-Pex The OTRYapViewTest testViewHandlerUpdate test now fails after merging this PR. Do you think you could take a look at why it's failing?

N-Pex:
@chrisballinger The AllBuddies view does not include people not on your roster, so the test requires a "buddy.trustLevel = .roster" at line 71 in OTRYapViewTest.swift. Alternatively, we could make sure that AllBuddies returns ALL buddy objects (regardless of trust level) and instead add a RosterBuddies view?

chrisballinger:
@N-Pex Thanks for looking into that! I'll modify just the test bc I think we shouldn't touch the yap views until we figure out the launch performance regression.

PR #929, opened 2018-01-11
Allow retry of failed downloads

N-Pex:
Also, show clock while waiting for read receipt. Or is there a reason why we didn't do this?


chrisballinger:
On upstream the flow is:

1. Clock - when message is in queue waiting to be sent to network
2. Nothing - Message has been "sent"
3. Check - Message receipt received

There are a few clients that don't support read receipts so the clock doesn't make sense to show if the message was sent to the network. Clock shouldn't be shown unless message hasn't left the device.

Issue #917, opened 2018-01-03
Search Bar Disappears/hide after clicking on search bar in OTRComposeViewController


kaleemozitech:
Search is working fine but when user tabs on search bar, he can't see what text is in search bar

ethylmorphine:
Confirm. Actually, this bug is there for a very long time.

chrisballinger:
I'm pretty sure it's new as of iOS 11

kaleemozitech:
I am not sure but I am seeing this from a long time

PR #934, opened 2018-01-23
Fix for issue #917, fixed wrong size after device turn.

tladesignz:


ChatSecure/Classes/View Controllers/OTRComposeViewController.m
```diff
@@ -157,6 +157,15 @@ - (void)viewWillAppear:(BOOL)animated
...
 
+- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
+{
+    // Resize UISearchBar manually - it doesn't do it on its own on device turn.
+    self.searchController.searchBar.frame = CGRectMake(0, 0, size.width, size.height);
+    [self.searchController.searchBar sizeToFit];
+
+    [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
+}
+
```
chrisballinger: I'm guessing this could be solved by adding the proper constraints elsewhere, but this solution is probably less brittle given the wonkiness of UISearchController's searchBar when embedded in views other than tableHeaderViews.

PR #935, opened 2018-01-23
Move removal of old objects into thread, after extension registration

N-Pex:


N-Pex:
@chrisballinger I can't explain exactly why the deletion of subscription objects code would hang, but moving the code to the extension registration code seems to fix the problem at least.

The last commit was pretty hairy. I noticed that sometimes, like 1 in 20, when I started the app the conversations list would populate - then unpopulate - leaving the list empty! Turns out that dependent views were not registered in the correct order. This could definitely have caused weird startup problems with empty views.

ChatSecure/Classes/Controllers/OTRDatabaseManager.m
```diff
@@ -214,6 +215,11 @@ - (BOOL)setupYapDatabaseWithName:(NSString *)name
         YapDatabaseSearchResultsView *searchResultsView = [[YapDatabaseSearchResultsView alloc] initWithFullTextSearchName:FTSName parentViewName:AllBuddiesName versionTag:nil options:nil];
         NSString* viewName = [YapDatabaseConstants extensionName:DatabaseExtensionNameBuddySearchResultsViewName];
         [self.database registerExtension:searchResultsView withName:viewName];
+        
+        // Remove old unused objects
+        [self.readWriteDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction * _Nonnull transaction) {
+            [transaction removeAllObjectsInCollection:OTRXMPPPresenceSubscriptionRequest.collection];
+        }];
```
chrisballinger: We could also do a read-only check for the numberOfObjectsInCollection OTRXMPPPresenceSubscriptionRequest.collection before issuing the readwrite. Not sure why this takes so long because I figured it would be an O(1) operation with minimal side effects, possibly it triggers some relationship recalculations.

chrisballinger:
Thanks for looking into that! The out-of-order extension registration definitely would cause issues as well.

PR #940, opened 2018-01-25
Group fixes

N-Pex:
Store buddyUniqueId on room messages (if we have it)


ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -467,6 +468,22 @@ - (void)setThreadKey:(NSString *)key collection:(NSString *)collection
         }];
     }
```
chrisballinger: Add comment here explaining `serverCheckUpdateNotificationObject`

PR #941, opened 2018-01-29
Occupant deduplication

N-Pex:
I think we have a few things to discuss here, but wanted to get the code up there at least.

- [ ] The secondary index code is a real hack now, looking up a jid in an array. Currently done by wrapping each jid in \t and doing an SQL LIKE query. What's a good fix?
- [ ] enumerateRoomOccupants, which uses the secondary index, was unused so I commented it out. Remove altogether?
- [ ] Is there a good way to know if MAM support is available when joining a room, i.e. do we know if the call to "fetchHistory" will succeed? In that case we should set the "traditional" history element to maxChars = 0 when joining so that we don't have to deal with duplicate incoming messages. Also, we prefer to get them from MAM because we get the desired "item" element so we can map the messages to actual sender buddies (in the "traditional" history we can only succeed with this if the sender is available in the room, i.e. an "occupant" in the XMPP sense. If they are offline we can't).

### DIFF
--- a/ChatSecure/Classes/Controllers/XMPP/Storage/RoomStorage.swift
+++ b/ChatSecure/Classes/Controllers/XMPP/Storage/RoomStorage.swift
@@ -148,9 +148,9 @@ extension RoomStorage: XMPPRoomStorage {
             let role = item.attributeStringValue(forName: "role") ?? ""
             let affiliation = item.attributeStringValue(forName: "affiliation") ?? ""
             if presence.presenceType == .unavailable {
-                occupant.removeJid(presenceJID.full)
+                occupant.removeJid(presenceJID)
             } else {
-                occupant.addJid(presenceJID.full)
+                occupant.addJid(presenceJID)
             }
             occupant.roomName = presenceJID.resource
             occupant.role = RoomOccupantRole(stringValue: role)
--- a/ChatSecure/Classes/Controllers/XMPP/Storage/RoomStorage.swift
+++ b/ChatSecure/Classes/Controllers/XMPP/Storage/RoomStorage.swift
@@ -90,9 +90,19 @@ import YapDatabase
             let message = OTRXMPPRoomMessage(message: xmppMessage, delayed: delayed, room: room)
             message.originId = originId
             message.stanzaId = stanzaId
-            
-            if let sender = message.senderJID, let senderJid = XMPPJID(string: sender), let roomJid = room.roomJID, let occupant = OTRXMPPRoomOccupant.occupant(jid: senderJid, realJID: nil, roomJID: roomJid, accountId: account.uniqueId, createIfNeeded: false, transaction: transaction) {
-                message.buddyUniqueId = occupant.buddyUniqueId
+
+            // Try to find buddy of sender. We might get an muc#user item element from where we can pull the real jid of the sender, else we try by message.senderJID.
+            if let x = xmppMessage.element(forName: "x", xmlns: XMPPMUCUserNamespace), let item = x.element(forName: "item"), let jidString = item.attribute(forName: "jid")?.stringValue, let jid = XMPPJID(string: jidString) {
+                if let buddy = OTRXMPPBuddy.fetchBuddy(jid: jid, accountUniqueId: account.uniqueId, transaction: transaction) {
+                    message.buddyUniqueId = buddy.uniqueId
+                }
+                // Is this from us?
+                if let accountJid = account.bareJID, jid.bareJID.isEqual(to: accountJid) {
+                    message.state = .sent
+                }
+            }
+            if message.buddyUniqueId == nil, let senderJidString = message.senderJID, let senderJid = XMPPJID(string: senderJidString), let roomJid = room.roomJID, let occupant = OTRXMPPRoomOccupant.occupant(jid: senderJid, realJID: senderJid, roomJID: roomJid, accountId: account.uniqueId, createIfNeeded: false, transaction: transaction), let buddy = occupant.buddy(with: transaction) {
+                message.buddyUniqueId = buddy.uniqueId
             }
             
             room.lastRoomMessageId = message.uniqueId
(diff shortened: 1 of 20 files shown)


ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
```diff
@@ -93,10 +93,12 @@ open class OTRXMPPRoomOccupant: OTRYapDatabaseObject, YapDatabaseRelationshipNod
...
     
-    @objc open var available = false
+    @objc open var available:Bool {
+        return (jids?.count ?? 0) > 0
+    }
     
-    /** This is the JID of the participant as it's known in the room i.e. baseball_chat@conference.dukgo.com/user123 */
-    @objc open var jid:String?
+    /** This is all JIDs of the participant as it's known in the room i.e. baseball_chat@conference.dukgo.com/user123 */
+    @objc open var jids:[String]?
```
chrisballinger: I think it would be better to only expose a computed `Set<XMPPJID>` instead of an array of strings, but keep this variable as private internal storage. For example:

```swift
    @objc private var _jids: [String]?
    
    @objc open var jids: Set<XMPPJID> {
        get {
            let validJids = _jids?.flatMap({ (jidStr) -> XMPPJID? in
                XMPPJID(string: jidStr)
            })
            return Set(validJids ?? [])
        }
        set {
            _jids = newValue.map { (jid) -> String in
                jid.full
            }
        }
    }
```

ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
```diff
@@ -161,8 +176,9 @@ public extension OTRXMPPRoomOccupant {
         
         var parameters: [String] = [roomUniqueId]
         var queryString = "Where \(RoomOccupantIndexColumnName.roomUniqueId) == ? AND ("
-        parameters.append(jid.full)
-        queryString.append("\(RoomOccupantIndexColumnName.jid) == ?")
+        // We build the secondary index with appended \0 to avoid matching wrong jids, so add a matching \0s here.
+        parameters.append("%\t\(jid.full)\t%")
+        queryString.append("\(RoomOccupantIndexColumnName.jids) LIKE ?")
```
chrisballinger: Although it's probably not ideal, if it works as intended I think we should move forward for now until we think of a better/faster solution.

ChatSecure/Classes/Model/Yap Storage/OTRXMPPRoomOccupant.swift
```diff
@@ -206,7 +225,28 @@ public extension OTRXMPPRoomOccupant {
...
             occupant?.buddyUniqueId = buddy.uniqueId
         }
-        
         return occupant
     }
 }
+
+// Extension for adding/removing jids from the jids array
+public extension OTRXMPPRoomOccupant {
+    @objc public func addJid(_ jid:String) {
```
chrisballinger: See my commentary above about using a `Set<XMPPJID>` instead

chrisballinger:
> Is there a good way to know if MAM support is available when joining a room, i.e. do we know if the call to "fetchHistory" will succeed? In that case we should set the "traditional" history element to maxChars = 0 when joining so that we don't have to deal with duplicate incoming messages. Also, we prefer to get them from MAM because we get the desired "item" element so we can map the messages to actual sender buddies (in the "traditional" history we can only succeed with this if the sender is available in the room, i.e. an "occupant" in the XMPP sense. If they are offline we can't).

There is, but the capabilities storage is only kept in memory which limits its usefulness. We would need to write a Yap CapabilitiesStorage class.... :-\

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -1203,11 +1199,13 @@ - (void)updateRangeOptions:(BOOL)reset
...
+    __block NSUInteger shownCount;
+    __block NSUInteger totalCount;
     [self.readOnlyDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
-        NSUInteger shownCount = [self.viewHandler.mappings numberOfItemsInGroup:self.threadKey];
-        NSUInteger totalCount = [[transaction ext:OTRFilteredChatDatabaseViewExtensionName] numberOfItemsInGroup:self.threadKey];
-        [self setShowLoadEarlierMessagesHeader:shownCount < totalCount];
+        shownCount = [self.viewHandler.mappings numberOfItemsInGroup:self.threadKey];
+        totalCount = [[transaction ext:OTRFilteredChatDatabaseViewExtensionName] numberOfItemsInGroup:self.threadKey];
     }];
+    [self setShowLoadEarlierMessagesHeader:shownCount < totalCount];
```
chrisballinger: Good catch!

N-Pex:
Pretty happy with the last two commits here, should fix some strangeness happening. I think I am done adding stuff to this PR now. XMPPFramework needs to be updated for all this to build, but unsure what branch to use, I saw you used xep_0319 branch at the moment(?)

chrisballinger:
We can use the master branch of upstream XMPPFramework now, once I merge your other PR there. Really excited about the results from your deep dive uncovering the session weirdness!

ChatSecure/Classes/Controllers/XMPP/Storage/RoomStorage.swift
```diff
@@ -103,6 +103,10 @@ import YapDatabase
             }
             if message.buddyUniqueId == nil, let senderJidString = message.senderJID, let senderJid = XMPPJID(string: senderJidString), let roomJid = room.roomJID, let occupant = OTRXMPPRoomOccupant.occupant(jid: senderJid, realJID: senderJid, roomJID: roomJid, accountId: account.uniqueId, createIfNeeded: false, transaction: transaction), let buddy = occupant.buddy(with: transaction) {
```
chrisballinger: @N-Pex Minor nitpick (not blocking this merge) but in the future please insert newlines between the if conditions when they don't fit on a single line. Xcode will automatically line them up if you hit enter after the ", "

N-Pex: Ok!

PR #946, opened 2018-02-01
Knock button would immediately be overwritten?

N-Pex:
Maybe just my ignorance about how the knock button works, but it seems to me that the knock button would be replaced by mic/keyboard on all servers that support http uploads? Maybe this is intended behavior? I haven't been able to test the knock functionality, unsure exactly what would trigger it.


chrisballinger:
Knock button and all associated code can be removed! We no longer use Knocks/OTRDATA at all now that we have XEP-0357 and XEP-0363. :)

N-Pex:
Haha, excellent, thanks, closing this =)

PR #948, opened 2018-02-05
Wait for authenticated before start sending IQ stanzas.

N-Pex:
In about 1 in 10 starts of the app I noticed didNotAuthenticate was
called, because IQ stanza was sent (and the response taken for an auth
response error) before auth was done.


chrisballinger:
I saw this bug before and thought I attempted to fix it further up the stack so the query itself wouldn't be sent until we were authenticated

PR #952, opened 2018-02-07
Cancel and resume downloads

N-Pex:
Resume interrupted downloads on connect. Show a "Cancel" option in message view controller when avatar is tapped on active download.


ChatSecure/Classes/Controllers/FileTransferManager.swift
```diff
@@ -140,6 +153,21 @@ public class FileTransferManager: NSObject, OTRServerCapabilitiesDelegate {
     
     // MARK: - Public Methods
     
+    // Resume downloads, i.e. look for media items that are partially downloaded and retry getting them. TODO - use ranges
+    @objc public func resumeDownloads() {
+        connection.asyncRead { (transaction) in
+            transaction.enumerateKeysAndObjects(inCollection: OTRMediaItem.collection, using: { (key, object, stop) -> Void in
```
chrisballinger: Enumerating every object in the collection will be extremely slow, and this will block the global write queue, which may have cascading consequences where we do blocking reads on a write connection. I think that a SecondaryIndex for OTRMediaItems should be used on `.transferProgress` before this can be merged.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -769,6 +769,24 @@ - (nonnull UIAlertAction *)cancleAction {
         [actions addObject:resendAction];
     }
     
+    // If we are currently downloading, allow us to cancel
+    if([[message messageMediaItemKey] length] > 0 && [message conformsToProtocol:@protocol(OTRDownloadMessage)] && message.messageError == nil) {
```
chrisballinger: Please factor this out into a method that returns a nullable UIAlertAction

PR #953, opened 2018-02-07
WIP: More occupant fixes

N-Pex:
Sigh, now after these changes, hanging on to the jid:s, I wonder if we maybe should have sticked to having XMPPRoomOccupant as an occupant and not a container object...  Ah well. Consider this Work In Progress, still stuff to do here, but mostly wanted to see how you like the idea of storing online/offline in the buddy cache (which nowadays is a bit of a misnomer)?

Also in this PR - delay fetching the history until we are done getting room members, admins and owners. Currently really only works in theory since we 1. Use old-school history elements and 2. Fetch MAM history for everything when we connect to the server, so the results will come "too early".

### DIFF
--- a/ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomManager.m
+++ b/ChatSecure/Classes/Controllers/XMPP/OTRXMPPRoomManager.m
@@ -159,7 +159,6 @@
         
         OTRXMPPRoomOccupant *occupant = [transaction objectForKey:edge.sourceKey inCollection:edge.sourceCollection];
         occupant.role = RoomOccupantRoleNone;
-        occupant.jids = nil;
         [occupant saveWithTransaction:transaction];
     }];
 }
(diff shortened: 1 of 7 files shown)


chrisballinger:
I think we should move back to RoomOccupants corresponding to exactly one roomJID/nickname `jid`, because I think that is the source of problems we are having with duplication. When you have the case where a RoomOccupant has no jids associated with it anymore, that object has no more context and will simply float around forever in the database, which will cause another occupant object to be created. I also think the `roomName` property should be replaced with a computed property of `jid.resource`.

chrisballinger:
I think I will manually revert the storing of multiple `jids` on the Occupant in favor of storing them in the runtime cache.

chrisballinger:
@N-Pex See this commit: 0a77d49d

In the `zom-more_occupant_fixes` branch I've merged this PR and manually reverted the JIDs Set. In the process I seem to have fixed the root problem of duplicate occupants. Basically occupants can co-exist in these states:

* Room JID only (anonymous rooms)
* Real JID only (occupant list)
* Both Room & Real JID (non-anon presence update)

We now do a lookup for either roomJID OR realJID, and then fill in the other values as we can. Occupants without either are invalid and are filtered out. Ownership of an Occupant by a "realJID" can be murky but basically we just overwrite existing occupants with the most recently seen realJID. This logic can probably be improved, but so far it has been working well.

chrisballinger:
Merged into master branch

PR #958, opened 2018-02-14
Don't hide, just disable "add friends" and "Change subject"...

N-Pex:
...If we have the permissions, but are not currently in the room (i.e.
offline).


chrisballinger:
Looks good. I do wish we had the time to rewrite the Occupants list to just be a standard table view, instead of weird stack view cells in the header...

PR #961, opened 2018-02-19
Add "add friend" functionality to unknown sender cell

N-Pex:


ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -2397,15 +2408,71 @@ - (nullable NSArray *)supplementaryViewsForCellAtIndexPath:(NSIndexPath *)indexP
     return nil;
 }
 
-- (void)populateUnknownSenderCell:(OTRMessagesUnknownSenderCell *)cell withMessage:(id<OTRMessageProtocol,JSQMessageData>)message {
+- (void)populateUnknownSenderCell:(OTRMessagesUnknownSenderCell *)cell withMessage:(id<OTRMessageProtocol,JSQMessageData>)message indexPath:(NSIndexPath *)indexPath {
```
chrisballinger: Please move this to an instance method on `OTRMessagesUnknownSenderCell`. Although it's kind of verbose and not exactly ideal, you should be able to pass in all of the dependencies individually.

We really should try to prevent adding any more lines to OTRMessagesViewController, and aim for net-negative over time.

ChatSecure/Classes/View Controllers/OTRMessagesViewController.m
```diff
@@ -2397,15 +2408,71 @@ - (nullable NSArray *)supplementaryViewsForCellAtIndexPath:(NSIndexPath *)indexP
...
+                __block OTRAccount *account = nil;
+                [strongSelf.readOnlyDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction * _Nonnull transaction) {
+                    account = [strongSelf accountWithTransaction:transaction];
+                }];
+                if (account) {
+                    OTRXMPPManager *manager = (OTRXMPPManager *)[[OTRProtocolManager sharedInstance] protocolForAccount:account];
+                    if (manager) {
+                        [manager addToRosterWithJID:senderJID displayName:senderDisplayName];
+                        dispatch_async(dispatch_get_main_queue(), ^{
+                            [self.collectionView reloadData];
```
chrisballinger: Can you reload just the affected cell at `indexPath`?

N-Pex:
A whole bunch of changes here, moving out supplementary view handling to a swift extension. A sad side effect of this are the two attached object properties. We can move those to be "real" properties in the .h and .m but I would rather keep them in the extension. What do you think?

ChatSecure/Classes/View Controllers/OTRMessagesViewController+SupplementaryViews.swift
```diff
@@ -0,0 +1,252 @@
...
+//  Copyright © 2018 Chris Ballinger. All rights reserved.
+//
+
+import Foundation
+import OTRAssets
+
+var OTRMessagesViewController_supplementaryViewsAssociatedObject: UInt8 = 0
+var OTRMessagesViewController_prototypeCellsAssociatedObject: UInt8 = 0
+
+extension OTRMessagesViewController: OTRMessagesCollectionViewFlowLayoutSupplementaryViewProtocol {
```
chrisballinger: Can you make this a standalone class conforming to OTRMessagesCollectionViewFlowLayoutSupplementaryViewProtocol and store it as a private member inside OTRMVC.m? You should be able to make it standalone if you just init it with DatabaseConnections and pass in the message at indexPath in supplementaryViewsForCellAtIndexPath. I'd prefer to avoid objc_getAssociatedObject :)

ChatSecure/Classes/View Controllers/OTRMessagesViewController.h
```diff
@@ -47,7 +48,7 @@
 - (nullable UIBarButtonItem *)rightBarButtonItem;
 
 - (void)infoButtonPressed:(nullable id)sender;
-
+- (void)newDeviceButtonPressed:(nonnull NSString *)buddyUniqueId;
```
chrisballinger: Can you add `sender:(id)sender`?

chrisballinger:
Looking a lot better! I think we should avoid objc_getAssociatedObject by making the extension a standalone class internal to OTRMVC, and calling it as needed. Should be easy with some minimal changes, and then you won't have to expose anything in the public header either.

chrisballinger:
@N-Pex I merged this into upstream's `zom-something_changed` and did the refactor myself, but it could use a little cleanup: https://github.com/ChatSecure/ChatSecure-iOS/commit/1649c46643685bb5163133fefb74cc367eaa9779

I also noticed there seems to be some issue rendering the avatar: 

screen shot 2018-02-27 at 4 14 07 pm

PR #981, opened 2018-04-03
4.3.0 Group fixes

afriedmanGlacier:
A few proposed fixes due to database or UI updates causing errors from being on the wrong thread. Also, in the OTRRoomOccupantsViewController, turning the notification switch on/off was causing didChangeNotificationSwitch to be called twice (and screwing up the setting) because of the refreshNotificationSwitch call at the end of that function. I don't think refreshNotificationSwitch is needed there, and commenting that line out made it work as expected for us.

The Notification switch issue might be a result of the issue mentioned here: https://stackoverflow.com/questions/39566361/uiswitch-seton-animated-does-not-work-as-document


ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -461,7 +461,9 @@ open class OTRRoomOccupantsViewController: UIViewController {
         if let vc = storyboard.instantiateInitialViewController() as? OTRComposeGroupViewController {
             vc.delegate = self
             vc.setExistingRoomOccupants(viewHandler: self.viewHandler, room: self.room)
-            self.navigationController?.pushViewController(vc, animated: true)
+            DispatchQueue.main.async {
+                self.navigationController?.pushViewController(vc, animated: true)
+            }
```
chrisballinger: If this is happening off of the main thread, all of the other UIKit code above should be put on the main thread as well. Proper fix will be to ensure that addMoreFriends isn't being called off the main queue from the call site. However, it looks like this code is only ever triggered from `didSelectHeaderCell` which is only called from `    public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath)` which is guaranteed to be on the main thread.

Under what conditions is `addMoreFriends` called off the main thread?

ChatSecure/Classes/View Controllers/OTRComposeGroupViewController.swift
```diff
@@ -224,7 +224,7 @@ open class OTRComposeGroupViewController: UIViewController, UICollectionViewDele
                     for row in 0..<mappings.numberOfItems(inSection: section) {
                         var buddy:OTRXMPPBuddy? = nil
                         if let roomOccupant = viewHandler.object(IndexPath(row: Int(row), section: Int(section))) as? OTRXMPPRoomOccupant {
-                            OTRDatabaseManager.shared.uiConnection?.read({ (transaction) in
+                            OTRDatabaseManager.shared.readConnection?.read({ (transaction) in
```
chrisballinger: Good catch. Although it's not enforced, the `uiConnection` shouldn't be used off the main thread.

ChatSecure/Classes/View Controllers/OTRRoomOccupantsViewController.swift
```diff
@@ -417,7 +417,7 @@ open class OTRRoomOccupantsViewController: UIViewController {
         connection?.readWrite({ (transaction) in
             room.save(with: transaction)
         })
-        refreshNotificationSwitch(notificationSwitch, room: room, animated:true)
+        //refreshNotificationSwitch(notificationSwitch, room: room, animated:true)
```
chrisballinger: I prefer not to leave stray commented/dead code hanging around. It looks like this can safely be removed though 👍

afriedmanGlacier: Yep, the only reason I left it was just in case it made it easier to verify. I'll remove it.

chrisballinger:
In the future, please split up PRs into discrete units if possible to prevent standalone fixes from blocking each other. For example I'd merge the `uiConnection` -> `readConnection` fix as is, but it's blocked on unrelated code review feedback.

Additionally if you do `git rebase -i origin/master` before pushing future PR branches, it allows you to clean up stuff like "Merge pull request #2 from ChatSecure/master". I'm not totally anal about it but it can help keep things tidy.

Thanks!

ChatSecure/Classes/View Controllers/OTRComposeGroupViewController.swift
afriedmanGlacier: Was causing YapDatabaseConnection -> readWithBlock -> NonMainThreadException for me

afriedmanGlacier:
I apologize for my ignorance. Not sure what I'm doing wrong with rebasing. Feel free to reject it altogether and I can start over and try again.

chrisballinger:
No need to close/reopen, but you'll need to force push your rebased PR branch since history is being rewritten. Here's a guide: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

You'll wanna `pick` the first commit and `squash` the rest.

afriedmanGlacier:
Not sure why there are still two commits, but it got rid of some. Git frustrates me, sigh

chrisballinger:
No worries, I'll just squash it on my end. Thanks!

PR #1072, opened 2018-12-04
Use a different security policy for validating certificates against CAs

mimi89999:


mimi89999:
In several days I will make another PR that will make certificate pinning optional.

chrisballinger:
@mimi89999 Thank you! I will take some time to test this and try to get a new release out shortly. 

I have been considering making cert pinning optional for a long time. As long as it's toggled on a per-account basis, and that the default behavior for accounts migrated from older versions is the same (e.g. keep legacy accounts pinned by default), your next PR should be good to merge.

mimi89999:
I was considering making certificate pinning a global setting as it is in Conversations and other applications. Why should it be per account?

chrisballinger:
My thoughts were that different accounts may have different security requirements. You might not be as concerned about a certificate change on a lower risk account than another one used for more sensitive communications.

Issue #810, opened 2017-07-05
File Transfer Not working - OMEMO


john4526:
After ChatSecure version 4.1 was pushed through the app store  my images cannot will send.

I see that it is now trying to send via OMEMO vs. OTR.  I don't know what is wrong.  I get either an error stating an SSL error occurred where an secure connection cannot be established or I get Chatsecure filetransfererror error 3.

I am using prosody on the server side.  Are there specific modules that are required to support this besides HTTP_upload?

futurealecks:
Are you using https for file transfer? If so, do you have a self signed cert?

john4526:
I am using letsencrypt for the server cert.  Do I need a cert for the http upload domain?  I have the http_upload setup in my prosody component section of my config.  Do I need https somewhere, like as a module?

john4526:
I have confirmed I am using http_load using https.  I can see it loading when I restart.  Now the error I get is solely the Chatsecure filetransfererror error 3 which I don't know where to start with.

chrisballinger:
Apologies, I thought I added "user friendly" error messages, they must not be showing up. One possibility is that the TLS version is too low and iOS is rejecting the connection. You should make sure your Prosody (0.9.12) and OpenSSL (1.1.0f) versions are the latest (beyond the Debian version).

mimi89999:
When I send a file from CS on iOS to Conversations on Android, HTTP upload works fine. When I do the other way round, without OMEMO I just see s link in CS and with OMEMO I just see an empty message. I am on stable 4.1.0.

chrisballinger:
Our link detection code isn't ideal for https:// scheme, and will require a
"filename" like `http://example.com/file.jpg`, or an image/video/audio mime
type. If you are using http:// scheme it will never show a preview because
of Apple's ATS. When using OMEMO it shouldn't be empty or blank... possibly
an unsupported format or too large?

mimi89999:
@chrisballinger The format is jpg, so it should be supported.
[...]

PR #1073, opened 2018-12-06
Fix displaying of encrypted attachments

mimi89999:
If the file is encrypted, the server might not know its mime type. If that's the case, it will set contenttype to application/octet-stream. If the file is encrypted and this mime type is advertised by the server, we must guess the mime type from the extension


mimi89999:
It fixes this issue: https://github.com/ChatSecure/ChatSecure-iOS/issues/810#issuecomment-313469521

chrisballinger:
Is it worth always guessing from the file extension for `application/octet-stream`, even if the file is not encrypted?

PR #1101, opened 2019-03-29
Show occupant nickname in occupant view for anonymous conferences

mimi89999:


mimi89999:
There is a memory leak when scrolling through the member list and opening and closing the occupant view of an anonymous MUC. I was searching for it, but couldn't find it.

chrisballinger:
thanks! this fixes the leak?

mimi89999:
No. It doesn't. I can't find it.

PR #1117, opened 2019-05-10
Changed certificate pinning to be optional

mimi89999:


chrisballinger:
The ideal approach would disable the cert pinning module entirely, I have a work in progress here: https://github.com/ChatSecure/ChatSecure-iOS/commit/e4a46fd74e479943e4f3a681140ea3be9f7c84dd

However there are some issues with my approach, namely the cert pinning dialog no longer shows up consistently because (I think) of a race condition between `lastConnectionError` and `loginStatus`.

schmittlauch:
Just want to confirm that in times of Let'sEncrypt certificates with short validity periods, these certificate re-verification warnings can be highly irritating and might train users to just click ok on everything.

ChatSecure.xcodeproj/project.pbxproj
```diff
@@ -476,6 +476,8 @@
 		D9A429D01F31169F00BD2545 /* UIAlertController+ChatSecure.swift in Sources */ = {isa = PBXBuildFile; fileRef = D9A429CF1F31169F00BD2545 /* UIAlertController+ChatSecure.swift */; };
 		D9A7756F1E43F8A200027864 /* ProxyXMPPStream.h in Headers */ = {isa = PBXBuildFile; fileRef = D9A7756D1E43F8A200027864 /* ProxyXMPPStream.h */; };
 		D9A775701E43F8A200027864 /* ProxyXMPPStream.m in Sources */ = {isa = PBXBuildFile; fileRef = D9A7756E1E43F8A200027864 /* ProxyXMPPStream.m */; };
+		D9A7BCE71E4554E200888A8E /* OTRXMPPStream.h in Headers */ = {isa = PBXBuildFile; fileRef = D9A7BCE51E4554E200888A8E /* OTRXMPPStream.h */; };
```
mostafa2009k, line 479: happy

mimi89999:
@chrisballinger Could you please finish implementing this? Currently I'm not able to work on OS X. I think that it's very important to have this feature added since many XMPP servers are using LE certificates and it's very annoying for users to click through this warning every 2 months. I could even pay to get this added.

mimi89999:
@chrisballinger Could you please implement this?

PR #1180, opened 2020-02-16
Accept 12-byte IV for OMEMO media

### LINKED ISSUE #1171: Support 12 byte IV for OMEMO and aesgcm://
Conversations (and others) will very soon start sending 12 byte IVs for both OMEMO messages as well as the aesgcm:// uri scheme in order to better support libraries on other plattforms that only accept 12 bytes.

chrisballinger:
This should resolve the infinitely loading media messages sent by newer versions of Conversations.

Resolves #1171


Neustradamus:
@chrisballinger: Warning, it is not complete.

Linked to:
- https://github.com/anurodhp/Monal

chrisballinger:
@Neustradamus Can you elaborate?

Neustradamus:
@chrisballinger: I think that the best solution is to request help in the ticket entry :)

PR #1193, opened 2020-03-08
Enable Tor

drwdal:
You should allow Tor users to have the choice for anti censorship purposes.

### DIFF
--- a/OTRResources/Branding.plist
+++ b/OTRResources/Branding.plist
@@ -43,7 +43,7 @@
 	<key>AllowsDonation</key>
 	<true/>
 	<key>TorEnabled</key>
-	<false/>
+	<true/>
 	<key>AllowGroupOMEMO</key>
 	<true/>
 	<key>AllowDebugFileLogging</key>


licaon-kter:
You did read https://nitter.net/ChatSecure/status/1226639098040877057 right?

drwdal:
> 
> 
> You did read https://nitter.net/ChatSecure/status/1226639098040877057 right?

Yes, as well as the issue by @catgirlirl who I agree with, you should retain it in the main build for the sake of people who do not have iOS build environments and need to access XMPP over Tor

drwdal:
the benefit of Tor is that a user in a restrictive environment may still be allowed to access a mobile app store, eliminating support for Tor suddenly while said support remains viable puts those users at inconvenience if not at risk for a failure to find an alternative.

chrisballinger:
It may be re-enabled at a future date, but it will require additional time and resources to ensure that it is functional.

irlcatgirl:
I wonder if the United States FISA Courts send them an order to disable Tor and add a backdoor into the end to end encryption logic on the App Store builds. Would be cautious about continuing to use ChatSecure. This change seems sketchy at best 🤔

chrisballinger:
@irlcatgirl Lol that's not how any of this works. Fortunately the government still can't compel speech, including forcing companies to implement crypto backdoors.

The simple reality is that there aren't enough paying end users / recurring donations to allow me to work on this software full time. In order to support Mac Catalyst I had to do a refactoring of the Tor control socket code in CPAProxy, and was not able to test it sufficiently before the 5.0 release, so I have it disabled for now.

Want to see more active maintenance? Consider financially supporting the project and encouraging others to do the same.

drwdal:
> 
> 
> @irlcatgirl Lol that's not how any of this works. Fortunately the government still can't compel speech, including forcing companies to implement crypto backdoors.
> 
> The simple reality is that there aren't enough paying end users / recurring donations to allow me to work on this software full time. In order to support Mac Catalyst I had to do a refactoring of the Tor control socket code in CPAProxy, and was not able to test it sufficiently before the 5.0 release, so I have it disabled for now.
> 
> Want to see more active maintenance? Consider financially supporting the project and encouraging others to do the same.

Why not take the point of two interested parties who need this feature and would gladly use it if you put it in the main build, you can just add a disclaimer about reliability, socket control isn't a major issue presuming it's functional at any level, and performance isn't a great reason to exclude it.

@chrisballinger Apple can still limit your exposure on international stores in response to certain features being added/removed, it's easier to assume that would be the first reason you might be encouraged to remove support for Tor.

>Want to see more active maintenance? Consider financially supporting the project and encouraging others to do the same.

How much would it realistically cost for you to leave Tor support in the main build for people who don't want the privacy implications of TestFlight? It seems like it would be a minimal effort.

chrisballinger:
@drwdal The control socket not working is an issue because Tor support is not functional at all without it. The app is already not available in China due to Tor support, but that happened years ago during the great VPN ban wave.

Have you verified you're able to connect to Tor accounts and .onion services using your patch?

irlcatgirl:
If China were to outlaw end to end encryption and blocked apps which had it, would you remove OMEMO from ChatSecure to remain on the Chinese App Store?

chrisballinger:
@irlcatgirl What? It's already not in the Chinese App Store, and no, I wouldn't remove OMEMO.

irlcatgirl:
>  The app is already not available in China due to Tor support, but that happened years ago during the great VPN ban wave.

That quote gives the implication features that upset evil governments will be removed.

licaon-kter:
Monal and Siskin are on the China store with OMEMO, right?

chrisballinger:
@licaon-kter Yep, currently OMEMO is fine, but Tor is not. However any app can be banned at any time, for any reason, within their territory.

mimi89999:
@irlcatgirl Maybe the government sent you to break ChatSecure so that users switch to less secure alternatives?

@chrisballinger Can this PR be closed unless somebody proves that it's actually working?

irlcatgirl:
> @irlcatgirl Maybe the government sent you to break ChatSecure so that users switch to less secure alternatives?
> 
> @chrisballinger Can this PR be closed unless somebody proves that it's actually working?

Maybe I'm not actually a college student and work for the CIA 😜

